PR #20, opened 2023-09-28
fix: get_path_for implementation

### LINKED ISSUE #17: get_path_for is broken
The `get_path_for` method of `WebApiUrlFactory` doesn't work for me.

https://github.com/open-feature/python-sdk-contrib/blob/main/open_feature_contrib/providers/flagd/web_api_url_factory.py#L39-#L44

## Fix

```
def get_path_for(self, flag_type: FlagType):

        # This is complex. Explainer
        # 1. Look up the appropriate string from __mapping for the incoming flag_type
        #    for example, FlagType.STRING should match and return the string "get_string_path"
        #    if an invalid flag_type or something goes wrong, fallback to __default_mapping_key
        #    which is "_invalid_flag_type_method"
        # 2. Use that string eg. "get_string_path" with getattr to "make it available" as a method to call
        #    Notice that each string in __mapping has a corresponding method
        #    Thus method_to_call() is dynamic, based on flag_type, so it could be get_string_path() or get_float_path() etc.
        # 3. Invoke that method and assign the return to formatted_endpoint which should be the full URL to flagd
        #    eg. http://localhost:8013/schema.v1.Service/ResolveString
        method_to_call = getattr(self, self.__mapping.get(flag_type, WebApiUrlFactory.__default_mapping_key))
        formatted_endpoint = method_to_call()
        return formatted_endpoint
```

agardnerIT:
## This PR
- Fixes the implementation of `get_path_for` so it works

### Related Issues
Fixes #17

### DIFF
--- a/openfeature/contrib/providers/flagd/web_api_url_factory.py
+++ b/openfeature/contrib/providers/flagd/web_api_url_factory.py
@@ -37,11 +37,10 @@ class WebApiUrlFactory:
         return self._format_url(self.OBJECT)
 
     def get_path_for(self, flag_type: FlagType):
-        return self[
-            WebApiUrlFactory.__mapping.get(
-                flag_type, WebApiUrlFactory.__default_mapping_key
-            )
-        ]()
+        method_name = self.__mapping.get(
+            flag_type, WebApiUrlFactory.__default_mapping_key
+        )
+        return getattr(self, method_name)()
 
     def _format_url(self, path: str):
         return f"{self.root}/{path}"


odubajDT:
Hey @agardnerIT, thank you for the contribution! Can you please fix the unit tests?

federicobond (approved):

federicobond:
Hi @agardnerIT, I rebased and slightly edited your PR. Commit is still attributed to you.

PR #29, opened 2023-10-22
feat: implement grpc protocol version of FlagdProvider

federicobond:
## This PR

Implements a gRPC protocol version of the Flagd Provider. It has been tested manually but not extensively. There are some difference with the way it handles errors for which I couldn't find a definitive answer in the spec.

### Notes

 - Should we leave generated grpc code under version control or should we regenerate each time?

### Follow-up Tasks

 - [ ] Add appropriate tests, not sure what's the best approach, probably integration tests with a flagd docker instance
 - [ ] Add support for datetimes in evaluation context


beeme1mr:
Amazing, thanks! I'll take a look later tonight or tomorrow. @toddbaert should be able to help clarify how we handle generated code in other languages and any other implementation questions.

beeme1mr:
Hey @federicobond, sorry for the delay. I'll try and answer as many questions as I can. I may need @toddbaert to help with any remaining questions.

- We don't check in generated code in other repos. I'm assuming we could generate the code before releasing it to PIP.
- [Here](https://github.com/open-feature/js-sdk-contrib/blob/main/libs/providers/flagd/src/lib/service/grpc/grpc-service.ts#L274) are the error mappings used in the JS version. It looks pretty close what what you have. Did you have any specific questions about error handling?
- Please check how falsy values are handled in the generated code. We had issues in other languages where falsy values were unexpectedly omitted from the response.

federicobond:
Thank you for the review @beeme1mr!

Specifically what I want to know is whether I should raise an exception or return a FlagEvaluationDetails with each error. Java maps errors to exceptions. If I follow the spec correctly I believe it would be more idiomatic for Python to also raise an exception:

> In cases of abnormal execution, the provider MUST indicate an error using the idioms of the implementation language, with an associated error code and optional associated error message.

beeme1mr:
I would recommend raising an exception but either approach would work. This should only occur when there's an error, so the overhead of an exception is acceptable in my opinion.

federicobond:
Removed generated code, switched to raise exceptions and improved error handling.

.github/workflows/pullrequest.yml
```diff
@@ -55,6 +61,17 @@ jobs:
...
+        run: |
+          cd schemas
+          buf generate buf.build/open-feature/flagd --template protobuf/buf.gen.python.yaml
+          cd ../
+          rm -rf openfeature/contrib/providers/flagd/proto
+          mv proto/python openfeature/contrib/providers/flagd/proto
+          sed -i.bak 's/^from schema.v1 import/from . import/' openfeature/contrib/providers/flagd/proto/schema/v1/*.py
+          rm openfeature/contrib/providers/flagd/proto/schema/v1/*.bak
+          rmdir proto
+
```
beeme1mr, line 74: Could you just run `make grpc`?

federicobond: Yeah, I am going to update the pipeline soon though, so I wouldn't worry too much about it.

beeme1mr:
> I would recommend raising an exception but either approach would work. This should only occur when there's an error, so the overhead of an exception is acceptable in my opinion.

Another thing to consider is how the Python SDK decides to run the error hook. In most SDKs, this happens in a catch block. If that's the case in Python, you should throw.

toddbaert:
woah this is awesome! I will review today, tomorrow latest.

toddbaert:
> Should we leave generated grpc code under version control or should we regenerate each time?

My recommendation is to generate it each time, as you seem to have done.

I see no issues here, this is great progress and probably worthy of release on its own. In addition to what you've pointed out, there's a few other enhancements that could be made (I would be comfortable helping with them as well, if I can find the time):

- opening the update stream
  - use it in standard flagd[caching](https://flagd.dev/reference/specifications/rpc-providers/?h=caching#caching)
  - use it to fire change events (would require associated SDK support)
- unix socket support (optional)

openfeature/contrib/providers/flagd/provider.py
```diff
@@ -84,102 +97,105 @@ def resolve_boolean_details(
...
+            code = e.code()
+            message = f"received grpc status code {code}"
+
+            if code == grpc.StatusCode.NOT_FOUND:
+                raise FlagNotFoundError(message)
+            elif code == grpc.StatusCode.INVALID_ARGUMENT:
+                raise TypeMismatchError(message)
+            elif code == grpc.StatusCode.DATA_LOSS:
+                raise ParseError(message)
+            raise GeneralError(message)
```
toddbaert, line 181: These look right to me.

toddbaert (approved):
Approving from a flagd/spec perspective, but my Python is weak. I made some notes on further enhancements.

beeme1mr (approved):

PR #30, opened 2024-01-09
chore: replace black, flake8 and isort with ruff and ruff-format

gruebel:
## This PR

same as in open-feature/python-sdk#249

- merged both workflows `merge.yml` and `pullrequest.yml` into `build.yml`, like in the `python-sdk` project, they were almost identical, just the trigger was different
- also migrated from container based builds to the normal host ones by leveraging `setup-python` GHA, same as in the `python-sdk` project


beeme1mr:
Did you intend to remove CodeQL?

gruebel:
@beeme1mr it is still there, previously named `pullrequest.yml`, now `build.yml` file. Just need to expand at the bottom of the file, tries to hide itself 😄

beeme1mr:
🤦 my bad... the change looks good.

beeme1mr:
> @beeme1mr do you know what the required `build (python:3.8)` checks are? I don't see them anywhere defined, so not sure why they are hanging.

The branch protection rules need to be updated.

PR #36, opened 2024-01-25
ci: initial release please config

beeme1mr:
Signed-off-by: Michael Beemer <michael.beemer@dynatrace.com>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,76 @@
+name: Run Release Please
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+
+    # Release-please creates a PR that tracks all changes
+    steps:
+      - uses: google-github-actions/release-please-action@v3
+        id: release
+        with:
+          command: manifest
+          token: ${{secrets.RELEASE_PLEASE_ACTION_TOKEN}}
+          default-branch: main
+          signoff: "OpenFeature Bot <109696520+openfeaturebot@users.noreply.github.com>"
+    outputs:
+      release_created: ${{ steps.release.outputs.releases_created }}
+      all: ${{ toJSON(steps.release.outputs) }}
+      paths_released: ${{ steps.release.outputs.paths_released }}
+
+  # pypi-release:
+  #   needs: release-please
+  #   runs-on: ubuntu-latest
+  #   if: ${{ needs.release-please.outputs.release_created }}
+  #   environment: release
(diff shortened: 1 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -28,27 +28,20 @@ We use `pytest` for our unit testing, making use of `parametrized` to inject cas
 
 These are planned once the SDK has been stabilized and a Flagd provider implemented. At that point, we will utilize the [gherkin integration tests](https://github.com/open-feature/test-harness/blob/main/features/evaluation.feature) to validate against a live, seeded Flagd instance.
 
-### Packaging
```
beeme1mr, line 31: I didn't feel like we needed to maintain this list in the contrib doc. It will likely get out of date and provide minimal value.

.github/workflows/release.yml
```diff
@@ -0,0 +1,76 @@
...
+          command: manifest
+          token: ${{secrets.RELEASE_PLEASE_ACTION_TOKEN}}
+          default-branch: main
+          signoff: "OpenFeature Bot <109696520+openfeaturebot@users.noreply.github.com>"
+    outputs:
+      release_created: ${{ steps.release.outputs.releases_created }}
+      all: ${{ toJSON(steps.release.outputs) }}
+      paths_released: ${{ steps.release.outputs.paths_released }}
+
+  # pypi-release:
```
beeme1mr, line 26: This is the part I'm not sure about. I believe the rest should be good to go now.

federicobond: For building, we would need to cd into each package directory, run `hatch build`  (make sure that hatch is installed first) and that should generate a `dist/` directory with the sdist .tar.gz and the .whl files.

After that the `pypa/gh-action-pypi-publish` action should be able to take over (as long as it's pointed to each package directory)

gruebel: this sounds like a good plan. @beeme1mr we can merge this first and I can adjust it accordingly. Do we also want to directly leverage trusted publishing? If yes, you would need to configure it on `pypi` similar to `python-sdk` and for each package once 🙂

beeme1mr: We could either merge this or work together off this draft.

I'll look into setting up `pypi` like we did for the SDK, but I have a busy schedule this week, so I'm not sure when I'll get to it.

thisthat: Hey @beeme1mr, if you don't mind, I can take over this :)

beeme1mr:
closing in favor of #38

PR #37, opened 2024-02-07
chore: update ruff version to 0.2.1 and remove preview flag

federicobond:
Matches https://github.com/open-feature/python-sdk/pull/272

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,9 +1,10 @@
 default_stages: [commit]
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.1.11
+    rev: v0.2.1
     hooks:
       - id: ruff
+        args: [--fix]
       - id: ruff-format
 
   - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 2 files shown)


matthewelwell (commented):
@federicobond the change here looks sound to me but I'm confused by the workflows failure. I can't seen anything in this PR that would cause this but it seems that the same workflow passed on the [last run on the main branch](https://github.com/open-feature/python-sdk-contrib/actions/runs/7648101431) and [your last PR branch](https://github.com/open-feature/python-sdk-contrib/actions/runs/7647958981). Any ideas?

gruebel:
@matthewelwell is is probably a side-effect of the recently released version 0.4.2 with this change open-feature/python-sdk#268. The `AbstractProvider` class defines `get_provider_hooks()` as an abstract method with a default implementation, which somehow worked before, but now not anymore, which makes more sense for me

https://github.com/open-feature/python-sdk/blob/f9833ba75390008f1bb3c07149dc41be714a1f78/openfeature/provider/provider.py#L22-L24

either we add this method to the `FlagdProvider` class or we remove the `@abstractmethod` from the abstract class.

federicobond:
Yeah, weird that it worked before. I'm in favor of removing the `@abstractmethod` decorator from AbstractProvider, as this may have an impact in downstream providers.

federicobond:
This is now building correctly.

gruebel (approved):

Issue #44, opened 2024-02-22
ModuleNotFoundError when importing flagd provider


austinlparker:
Hi -- I'm getting the following error when initializing the FlagdProvider.

```
2024-02-22 13:17:41 ModuleNotFoundError: No module named 'openfeature.contrib.provider.flagd.proto'
2024-02-22 13:18:32 Traceback (most recent call last):
2024-02-22 13:18:32   File "/usr/src/app/recommendation_server.py", line 24, in <module>
2024-02-22 13:18:32     from openfeature.contrib.provider.flagd import FlagdProvider
2024-02-22 13:18:32   File "/usr/local/lib/python3.12/site-packages/openfeature/contrib/provider/flagd/__init__.py", line 1, in <module>
2024-02-22 13:18:32     from .provider import FlagdProvider
2024-02-22 13:18:32   File "/usr/local/lib/python3.12/site-packages/openfeature/contrib/provider/flagd/provider.py", line 43, in <module>
2024-02-22 13:18:32     from .proto.schema.v1 import schema_pb2, schema_pb2_grpc
2024-02-22 13:18:32 ModuleNotFoundError: No module named 'openfeature.contrib.provider.flagd.proto'
```

PR #45, opened 2024-02-22
fix: include proto file in build for openfeature-provider-flagd

federicobond:
This should fix https://github.com/open-feature/python-sdk-contrib/issues/44. hatch now generates the proto client files during the build stage.

We can probably find a cleaner solution later.


toddbaert (approved):

federicobond:
@toddbaert damn, hatch requires python 3.9+. I have to go now, but feel free to edit this PR

providers/openfeature-provider-flagd/pyproject.toml
```diff
@@ -28,13 +28,21 @@ Homepage = "https://github.com/open-feature/python-sdk-contrib"
 
 [tool.hatch]
 
+[[tool.hatch.build.hooks.build-scripts.scripts]]
+commands = [
+  "./scripts/gen_protos.sh",
```
gruebel, line 33:
```suggestion
  "buf generate buf.build/open-feature/flagd --template buf.gen.python.yaml",
  "sed -i.bak 's/^from schema.v1 import/from . import/' proto/schema/v1/*.py",
  "rm proto/schema/v1/*.bak",
```
then you don't need to change the `gen_protos.sh` file content

gruebel (commented):
I found a similar solution, but yours seems better 😄

gruebel:
> @toddbaert damn, hatch requires python 3.9+. I have to go now, but feel free to edit this PR

the plugin requires Python 3.9 😢

toddbaert:
> @toddbaert damn, hatch requires python 3.9+. I have to go now, but feel free to edit this PR

Bummer. Looks like 3.8 isn't eol until October... Not sure we should drop early, but that's an option

beeme1mr (approved):

gruebel (approved):

PR #62, opened 2024-02-24
feat: implement environment-variable based config

federicobond:


providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/config.py
```diff
@@ -0,0 +1,33 @@
...
+@dataclass
+class Config:
+    host: str = config("FLAGD_HOST", "localhost")
+    port: str = config("FLAGD_PORT", 8013, cast=int)
+    tls: bool = config("FLAGD_TLS", False, cast=str_to_bool)
+    timeout: int = 5
+
+    @classmethod
+    def build(cls, **kwargs):
+        return cls(**{k: v for k, v in kwargs.items() if v is not None})
```
gruebel, line 33: why not just use a normal class, then you don't need the `build()` method and you could remove the `config()` function and just use the inner function `env_or_default()` directly
```suggestion
class Config:
    def __init__(
        self,
        host: typing.Optional[str] = None,
        port: typing.Optional[int] = None,
        tls: typing.Optional[bool] = None,
        timeout: typing.Optional[int] = None,
    ):
        self.host = env_or_default("FLAGD_HOST", "localhost") if host is None else host
        self.port = env_or_default("FLAGD_PORT", 8013, cast=int) if port is None else port
        self.tls = env_or_default("FLAGD_TLS", False, cast=str_to_bool) if tls is None else tls
        self.timeout = 5 if timeout is None else timeout
```

federicobond: Yeah, I might have been a little too clever for my own good here, and got carried away.

providers/openfeature-provider-flagd/pyproject.toml
```diff
@@ -62,3 +62,6 @@ omit = [
   # exclude generated files
   "src/openfeature/contrib/provider/flagd/proto/*",
 ]
+
+[tool.ruff.lint.flake8-bugbear]
+extend-immutable-calls = ["openfeature.contrib.provider.flagd.config.config"]
```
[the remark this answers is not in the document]
federicobond: I wanted to scope the configuration change to the flagd package but apparently I ended up messing the whole ruff configuration.

gruebel (approved):
nice 🍻

PR #72, opened 2024-03-19
refactor: add mypy and fix typing issues

gruebel:
## This PR

- configuring `mypy` was not fun, because of the `protobuf` generated code 🙈

### DIFF
--- a/hooks/openfeature-hooks-opentelemetry/src/openfeature/contrib/hook/opentelemetry/__init__.py
+++ b/hooks/openfeature-hooks-opentelemetry/src/openfeature/contrib/hook/opentelemetry/__init__.py
@@ -16,7 +16,7 @@ class EventAttributes:
 class TracingHook(Hook):
     def after(
         self, hook_context: HookContext, details: FlagEvaluationDetails, hints: dict
-    ):
+    ) -> None:
         current_span = trace.get_current_span()
 
         variant = details.variant
@@ -38,6 +38,8 @@ class TracingHook(Hook):
 
         current_span.add_event(OTEL_EVENT_NAME, event_attributes)
 
-    def error(self, hook_context: HookContext, exception: Exception, hints: dict):
+    def error(
+        self, hook_context: HookContext, exception: Exception, hints: dict
+    ) -> None:
         current_span = trace.get_current_span()
         current_span.record_exception(exception)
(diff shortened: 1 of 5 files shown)


hooks/openfeature-hooks-opentelemetry/src/openfeature/contrib/hook/opentelemetry/__init__.py
```diff
@@ -38,6 +38,8 @@ def after(
 
         current_span.add_event(OTEL_EVENT_NAME, event_attributes)
 
-    def error(self, hook_context: HookContext, exception: Exception, hints: dict):
+    def error(
+        self, hook_context: HookContext, exception: Exception, hints: dict
```
federicobond, line 42: There is a `HookHInts` type in `openfeature.hook`, should we use that?

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/config.py
```diff
@@ -1,23 +1,27 @@
...
+T = typing.TypeVar("T")
+
 
 def str_to_bool(val: str) -> bool:
     return val.lower() == "true"
 
 
-def env_or_default(env_var, default, cast=None):
+def env_or_default(
+    env_var: str, default: T, cast: typing.Optional[typing.Callable] = None
```
federicobond, line 12: We could type the Callable to avoid the type error, but this is an internal function so it's fine for me either way.

federicobond (commented):
Awesome work @gruebel! I wasn't even aware we were not running mypy here.

hooks/openfeature-hooks-opentelemetry/src/openfeature/contrib/hook/opentelemetry/__init__.py
gruebel: good point

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/config.py
gruebel: yeah, sure. can do 🙂

hooks/openfeature-hooks-opentelemetry/src/openfeature/contrib/hook/opentelemetry/__init__.py
gruebel: so, it looks like the `HookHints` is not released yet 😄 we can adjust it after releasing a new Python SDK version

federicobond (approved):

PR #74, opened 2024-03-30
feat: in-process offline flagd resolver

colebaileygit:
## This PR

- Adds in-memory resolver for the flagd provider which watches a local JSON file to retrieve flag configurations
- Implements fractional, semver, stringcomp custom operators
- Integrates and passes gherkin test suite

### Related Issues

https://github.com/open-feature/flagd/issues/1264

### Follow-up Tasks

- Add gRPC sync source option when using in-process provider
- Add schemas as submodule to verify flags against expected schema during parsing

### How to test

```bash
cd providers/openfeature-provider-flagd
hatch run test
```


beeme1mr:
Amazing, thank you! I'll review this as soon as possible.

federicobond:
This is great work @colebaileygit, thank you! Will review as soon as possible too.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/in_process.py
```diff
@@ -0,0 +1,120 @@
...
+            evaluation_context.targeting_key if evaluation_context else None
+        )
+        variant = jsonLogic(flag["targeting"], json_logic_context, self.OPERATORS)
+        if variant is None:
+            return FlagResolutionDetails(default, reason=Reason.DEFAULT)
+        if isinstance(variant, bool):
+            variant = str(variant).lower()
+
+        value = flag["variants"].get(variant)
+        # TODO: Check type matches
```
beeme1mr, line 112: Please check the type of the value. If it doesn't match, throw a type mismatch error.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/custom_ops.py
```diff
@@ -0,0 +1,115 @@
...
+    elif op == ">":
+        return v1 > v2  # type: ignore[no-any-return]
+    elif op == ">=":
+        return v1 >= v2  # type: ignore[no-any-return]
+    elif op == "^":
+        return v1.major == v2.major  # type: ignore[no-any-return]
+    elif op == "~":
+        return v1.major == v2.major and v1.minor == v2.minor  # type: ignore[no-any-return]
+    else:
+        raise ValueError(f"Op not supported by sem_ver: {op}")
```
beeme1mr, line 115: Do you know how throwing will behave here? In the JS implementation, we return `false` if none of the operators match.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/file_watcher.py
```diff
@@ -0,0 +1,59 @@
...
+    def refresh_file(self) -> None:
+        while True:
+            time.sleep(1)
+            logging.debug("checking for new flag store contents from file")
+            last_modified = os.path.getmtime(self.file_path)
+            if last_modified > self.last_modified:
+                self.load_data(last_modified)
+
+    def load_data(self, modified_time: typing.Optional[float] = None) -> None:
+        # TODO: error handling
```
beeme1mr, line 37: Please add error handling or create a follow-up issue.

beeme1mr (commented):
This looks great! I left a few comments but the logic matches other implementations.

I really appreciate you working on this. It's exciting to see the Python SDK and providers maturing.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/in_process.py
colebaileygit: Looks like the openfeature client does this already: https://github.com/open-feature/python-sdk/blob/main/openfeature/client.py#L434

Should we still add an explicit check to the provider?

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/custom_ops.py
colebaileygit: Looks like openfeature client catches exceptions well: https://github.com/open-feature/python-sdk/blob/main/openfeature/client.py#L357-L389

WDYT about returning  open feature `ParseError` instead for better error codes? https://github.com/open-feature/python-sdk/blob/main/openfeature/exception.py#L89 (would then look to do similar in other places if so)

beeme1mr: Ideally, this is caught by the JSON schema validator. If it's not, we try and never throw in JsonLogic and instead return a falsy value. This aligns with the out-of-the-box JsonSchema operators and provides consistency across implementations.

colebaileygit: Updated all the custom ops to return `None` and only log errors

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/file_watcher.py
colebaileygit: Put in some proper error handling for file reads, and also moved the logic to another class so it is run at parse time and not evaluation time. Feel free to suggest improvements to the new structure

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/file_watcher.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def shutdown(self) -> None:
+        pass
+
+    def get_flag(self, key: str) -> typing.Optional[Flag]:
+        return self.flag_data.get(key)
+
+    def refresh_file(self) -> None:
+        while True:
+            time.sleep(1)
```
federicobond, line 35: please extract this value to a constant or class field

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/file_watcher.py
```diff
@@ -0,0 +1,71 @@
...
+            logging.debug("checking for new flag store contents from file")
```
federicobond, line 36: please replace the root logger to a logger under the `openfeature` or `openfeature.contrib` name

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/file_watcher.py
```diff
@@ -0,0 +1,71 @@
...
+                else:
+                    data = json.load(file)
+
+                self.flag_data = self.parse_flags(data)
+                logging.debug(f"{self.flag_data=}")
+                self.provider.emit_provider_configuration_changed(
+                    ProviderEventDetails(flags_changed=list(self.flag_data.keys()))
+                )
+            self.last_modified = modified_time or os.path.getmtime(self.file_path)
+        except Exception:
```
federicobond, line 55: do we have a more specific set of exceptions we can handle here?

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/file_watcher.py
colebaileygit: I'd also prefer to have it configurable on init of provider, shouldn't be hard to add

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/file_watcher.py
colebaileygit: went with `openfeature.contrib` 👍

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/file_watcher.py
colebaileygit: I will hunt them down and add what I can find, what is the expected behavior for unexpected exceptions though?

I presume we would prefer the client and background thread does not crash as a rule so that apps can continue as usual with fallback behavior without downtime until the root cause is fixed

colebaileygit: Added sensible exception set, but kept the catch-all case in addition. Found [similar approach in openfeature](https://github.com/open-feature/python-sdk/blob/main/openfeature/client.py#L357-L389) for reference so I find this consistent

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/file_watcher.py
colebaileygit: I think it would also be a good idea to include the schema via submodule and parse the verify the data against that explicitly at parse time

beeme1mr: Do you want to include that in the PR or should we create a follow up issue?

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/file_watcher.py
beeme1mr: Great, thanks. I'll let @federicobond resolve the conversation if he agrees but the change makes sense to me.

beeme1mr:
Hey @colebaileygit, would you be interested in joining the org? It makes it easier to coordinate on issues and is an initial step towards becoming a maintainer if that's a goal. I would be happy to sponsor you, but there's certainly no obligation.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/file_watcher.py
federicobond: Sure! I worry that unexpected exceptions may mask programmer errors here but I realize that I am looking at the problem from the perspective of an application developer.

In a library, the best DX may be to just swallow these kinds of errors.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/file_watcher.py
colebaileygit: Follow-up issue sounds good, not sure how simple it would be to introduce

beeme1mr (approved):

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/file_watcher.py
colebaileygit: Added this as configuration to provider itself which is injected to the file watcher 👍

colebaileygit:
Had a small battle with CI, but the test suite should be robust now finally 🎉 
As a bonus I made it so the pre-commit hook evaluates using python 3.8 instead of your version so that we can catch those errors offline too

beeme1mr:
I'll give @federicobond a chance to review since he has much more Python experience but it looks good to me.

colebaileygit:
Shall I add this small change into this PR, or keep the two separate since the fractional update is a WIP breaking change? https://github.com/colebaileygit/open-feature-python-sdk-contrib/pull/1/files

Now that test-harness is merged this seems cleaner to include

colebaileygit:
@federicobond @beeme1mr Anything else pending? We are starting to use this internally so being able to have an official versioned release would be amazing.

The CI error looks unrelated to the PR and some issue with codecov, so don't know what I can do to solve it unfortunately

beeme1mr:
I'll review it today and schedule a release if everything looks good.

Issue #83, opened 2024-04-27
[flagd-process] gRPC sync as flagd data source


colebaileygit:
In process evaluator can only read flagd data from files. Implementing gRPC sync would allow connecting to flagd sidecar, flagd proxy, or any other service that implements the same protocol

aepfli:
@toddbaert i think we can close this issue as it was done with #104

PR #84, opened 2024-04-27
feat: add grpc sync flag store

colebaileygit:
## This PR

- adds gRPC sync source for flagd in-process evaluator

### Related Issues

Fixes https://github.com/open-feature/python-sdk-contrib/issues/83


providers/openfeature-provider-flagd/docker-compose.yaml
```diff
@@ -0,0 +1,25 @@
...
+      context: test-harness
+      dockerfile: sync/Dockerfile
+    ports:
+      - 9090:9090
+  flagd-sync-unstable:
+    build:
+      context: test-harness
+      dockerfile: sync/Dockerfile.unstable
+    ports:
+      - 9091:9090
```
federicobond, line 25: Let's make sure this file is excluded from published package. We can do so from the `pyproject.toml`

providers/openfeature-provider-flagd/pyproject.toml
```diff
@@ -69,3 +69,8 @@ omit = [
   "src/openfeature/contrib/provider/flagd/proto/*",
   "tests/**",
 ]
+
+[tool.pytest.ini_options]
+addopts = [
+    "--import-mode=importlib",
+]
```
federicobond, line 76: What is this setting for?

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/grpc_watcher.py
```diff
@@ -0,0 +1,92 @@
...
+import typing
+
+import grpc
+
+from openfeature.event import ProviderEventDetails
+from openfeature.exception import ParseError
+from openfeature.provider.provider import AbstractProvider
+
+from ....config import Config
+from ....proto.flagd.sync.v1 import sync_pb2, sync_pb2_grpc
```
federicobond, line 14: For more than one level up, an absolute import is much easier to read and to update if necessary.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/grpc_watcher.py
```diff
@@ -0,0 +1,92 @@
...
+from ..flags import Flag, FlagStore
+
+logger = logging.getLogger("openfeature.contrib")
+
+
+class GrpcWatcherFlagStore(FlagStore):
```
federicobond, line 20: Might be a good idea to split the store from the connector as the Java provider does, to better separate the responsibilities of each.

federicobond (commented):
Awesome work @colebaileygit! I left a few comments and questions merely to help push this forward, without being exhaustive. I know this is still in draft.

providers/openfeature-provider-flagd/docker-compose.yaml
colebaileygit: Good point. Checking this now it also includes `tests` and `test-harness` etc. Should we simplify to only include src?
```
[tool.hatch.build.targets.sdist]
include = [
  "src",
]
```

providers/openfeature-provider-flagd/pyproject.toml
colebaileygit: Was to prevent errors with test files with same name in different folders, but it is not needed and according to docs it isn't clearly preferred over the default, so we can remove this: https://docs.pytest.org/en/7.1.x/explanation/pythonpath.html

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/grpc_watcher.py
colebaileygit: I was getting some mypy errors when using absolute import, need to check how to fix those:
```
providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/grpc_watcher.py:10: error:
Cannot find implementation or library stub for module named
"openfeature.contrib.provider.flagd.proto.flagd.sync.v1"  [import-not-found]
    from openfeature.contrib.provider.flagd.proto.flagd.sync.v1 import (
    ^
```

providers/openfeature-provider-flagd/docker-compose.yaml
federicobond: I'd have to check the generated build to confirm. If you can please make this change in a separate PR.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/grpc_watcher.py
colebaileygit: Couldn't figure this out in a simple way, seems related to using pre-commit at root of a "mono-repo" but I don't see any clear resources of how to structure it better. Maybe this is a follow-up issue?

federicobond: This [approach](https://github.com/pre-commit/pre-commit/issues/466#issuecomment-274282684) that may work as long as the number of packages remains manageable.

.github/workflows/build.yml
```diff
@@ -27,6 +27,28 @@ jobs:
           - "hooks/openfeature-hooks-opentelemetry"
           - "providers/openfeature-provider-flagd"
 
+    services:
```
beeme1mr, line 30: This is fine (and what we do in other contribs repos), but I'm wondering if we should try to leverage the docker-compose so we only have to bump the versions in one place. [Here's a GitHub discussion](https://github.com/orgs/community/discussions/47105#discussioncomment-4950570) about how it could be configured. I'll let you decide if you think it's worth it, but I figured I would mention it. 😄

providers/openfeature-provider-flagd/README.md
```diff
@@ -36,12 +49,17 @@ api.set_provider(FlagdProvider(
 
 The default options can be defined in the FlagdProvider constructor.
 
-| Option name    | Type & Values | Default   |
-|----------------|---------------|-----------|
-| host           | str           | localhost |
-| port           | int           | 8013      |
-| schema         | str           | http      |
-| timeout        | int           | 2         |
+| Option name                   | Type & Values | Default   |
```
beeme1mr, line 52: Can you set these properties via environment variables?

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/file_watcher.py
```diff
@@ -0,0 +1,96 @@
...
+        flag_store: FlagStore,
+        poll_interval_seconds: float = 1.0,
+    ):
+        self.file_path = file_path
+        self.provider = provider
+        self.poll_interval_seconds = poll_interval_seconds
+
+        self.last_modified = 0.0
+        self.flag_store = flag_store
+        self.emit_ready = False
```
Kavindu-Dodan, line 35: minor - This variable name confused me with ready event. Shall we rename this to something like `emit_on_success` ?

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/in_process.py
```diff
@@ -27,18 +30,23 @@ class InProcessResolver:
...
+                self.flag_store,
+                self.config.offline_poll_interval_seconds,
             )
-        self.flag_store = FileWatcherFlagStore(
-            self.config.offline_flag_source_path,
-            self.provider,
-            self.config.offline_poll_interval_seconds,
+            if self.config.offline_flag_source_path
+            else GrpcWatcher(self.config, self.provider, self.flag_store)
         )
```
Kavindu-Dodan, line 43: Not a python expert, but does this if condition work correctly ?

Kavindu-Dodan (commented):
overal looks good 🙌  nice work and I left few comments, check them when you can

providers/openfeature-provider-flagd/README.md
Kavindu-Dodan: Good point, could be a follow-up to add env car support matching other providers (see java [1])

[1] - https://github.com/open-feature/java-sdk-contrib/tree/main/providers/flagd#configuration-options

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/file_watcher.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def initialize(self, evaluation_context: EvaluationContext) -> None:
+        self.active = True
+        self.thread = threading.Thread(target=self.refresh_file, daemon=True)
+        self.thread.start()
+
```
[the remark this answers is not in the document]
colebaileygit: I would actually say starting regardless is intentional here? For example first read might fail or timeout, but state can still be repaired in background

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/in_process.py
colebaileygit: Might be hard to read from the multi-line formatting, but this is python's ternary operator, e.g.:
```my_var = True if x > 3 else False```

Functionally what this does is look at the `offline_flag_source_path` to determine whether the user wants to watch file system or grpc. I'd say this can be done better if we extend with more options in future, but with just 2 seemed sufficient to me

File: providers/openfeature-provider-flagd/README.md (hunk 6 above)
Comment (colebaileygit) [reply providers/openfeature-provider-flagd/README.md]: Yes, for all except timeout for some reason. Let me know if you'd like to tweak naming for example changing `timeout` to `deadline`. I added the env vars to readme for visibility

File: providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/file_watcher.py (hunk 7 above)
Comment (colebaileygit) [reply providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/file_watcher.py:35]: I went back and forth on naming so agree. Should we be overly-explicit here?
`should_emit_ready_on_success`

File: providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/in_process.py (hunk 8 above)
Comment (Kavindu-Dodan) [reply providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/in_process.py:43]: ahh thanks, didn't know about this ternary operator :)

File: providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/file_watcher.py (hunk 7 above)
Comment (Kavindu-Dodan) [reply providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/file_watcher.py:35]: @colebaileygit I am fine with the proposal :) it's a nit but to make it easy to understand the logic

File: providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/file_watcher.py (hunk 9 above)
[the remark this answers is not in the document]
Comment (Kavindu-Dodan) [reply providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/file_watcher.py:41]: I think we shouldn't attempt to re-read the file if the initial read failed. Think about starting with an invalid file path. We should fail fast in this scenario and log the outcome so end-users know it failed. 

However, once we have the initialization, it's a different scenario. We should emit provider error and retry connection/read with back-offs.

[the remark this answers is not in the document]
Comment (colebaileygit) [reply providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/file_watcher.py:41]: I love failing fast but imagine the file you read from is updated periodically by some other job and there is a syntax error. Now your pods couldn't scale out and new ones would all crash.

Perhaps giving the user the option to raise or swallow errors on initialization is a good idea? How do we handle this in other languages?

I'd be more worried about this behavior for something like grpc, http, k8s sources though, since those can be inherently unreliable

File: providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/file_watcher.py [hunk 10]
```diff
@@ -0,0 +1,96 @@
...
+        self.provider = provider
+        self.poll_interval_seconds = poll_interval_seconds
+
+        self.last_modified = 0.0
+        self.flag_store = flag_store
+        self.should_emit_ready_on_success = False
+
+    def initialize(self, evaluation_context: EvaluationContext) -> None:
+        self.active = True
+        self.thread = threading.Thread(target=self.refresh_file, daemon=True)
```
federicobond, line 39: Why is daemon set to true in this thread?

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/flags.py
```diff
@@ -1,7 +1,42 @@
...
+                )
+            flags = json.loads(transposed)
+
+        if not isinstance(flags, dict):
+            raise ParseError("`flags` key of configuration must be a dictionary")
+        self.flags = {key: Flag.from_dict(key, data) for key, data in flags.items()}
+
+        self.provider.emit_provider_configuration_changed(
+            ProviderEventDetails(flags_changed=list(self.flags.keys()))
+        )
```
federicobond, line 39: I would prefer to treat `AbstractProvider.emit_<event>` methods as protected. Can we have the provider listen to store changes and emit these events itself, to better separate responsibilities?

This would have the benefit of decoupling the FlagStore from a provider, making it easier to test.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/flags.py
```diff
@@ -1,7 +1,42 @@
...
+    def update(self, flags_data: dict) -> None:
+        flags = flags_data.get("flags", {})
+        evaluators: typing.Optional[dict] = flags_data.get("$evaluators")
+        if evaluators:
+            transposed = json.dumps(flags)
+            for name, rule in evaluators.items():
+                transposed = re.sub(
+                    rf"{{\s*\"\$ref\":\s*\"{name}\"\s*}}", json.dumps(rule), transposed
+                )
+            flags = json.loads(transposed)
```
federicobond, line 31: We should include unit tests for this parsing.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/flags.py
```diff
@@ -32,11 +67,15 @@ def __post_init__(self) -> None:
...
+        if "defaultVariant" in data:
+            data["default_variant"] = data["defaultVariant"]
+            del data["defaultVariant"]
 
-        return flag
+        try:
+            flag = cls(key=key, **data)
+            return flag
+        except Exception as err:
+            raise ParseError from err
```
federicobond, line 78: Add unit tests for this method please.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/grpc_watcher.py
```diff
@@ -0,0 +1,103 @@
...
+                logger.debug("Setting up gRPC sync flags connection")
+                for flag_rsp in self.stub.SyncFlags(request):
+                    flag_str = flag_rsp.flag_configuration
+                    logger.debug(
+                        f"Received flag configuration - {abs(hash(flag_str)) % (10 ** 8)}"
+                    )
+                    self.flag_store.update(json.loads(flag_str))
+
+                    if not self.connected:
+                        self.provider.emit_provider_ready(
```
federicobond, line 72: Here too it I would suggest to decouple the connectors from providers, and send provider events only from inside the provider.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/file_watcher.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # Let this throw exceptions so that provider status is set correctly
+        try:
+            self._load_data()
+            self.should_emit_ready_on_success = True
+        except Exception as err:
+            raise ProviderNotReadyError from err
+
+    def shutdown(self) -> None:
+        self.active = False
```
federicobond, line 50: Shutting down and then initializing again can leave more than one thread running here, because `active` might not be checked until after the sleep expires.

federicobond: We have the same issue in both connector implementations. The current thread handling logic is a bit fragile, would it make sense to build a cancellable scheduled task manager and concentrate our efforts on making sure its implementation is robust? I'm happy to discuss how to approach it.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/file_watcher.py
federicobond: Let's add a name to all child threads for easier debugging. This could be `FlagdFileWatcherWorkerThread`

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/grpc_watcher.py
```diff
@@ -0,0 +1,103 @@
...
+        self.stub = sync_pb2_grpc.FlagSyncServiceStub(self.channel)
+        self.timeout = config.timeout
+        self.retry_backoff_seconds = config.retry_backoff_seconds
+        self.selector = config.selector
+
+        self.connected = False
+
+    def initialize(self, context: EvaluationContext) -> None:
+        self.active = True
+        self.thread = threading.Thread(target=self.sync_flags, daemon=True)
```
federicobond, line 40: We can name this one `FlagdGrpcSyncWorkerThread`

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/file_watcher.py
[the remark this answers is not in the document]
Kavindu-Dodan: Yeah that's a valid point. 

> How do we handle this in other languages?

Both Go & Java have fail fast strategy where if the initial read fails, provider initial fail, making evaluations to default. 

Don't take this as a blocker for this PR but this something to rethink @beeme1mr @toddbaert  any ideas ?

beeme1mr:
Hey @colebaileygit, could you please resolve all the conversations that have been addressed? It would be great to merge this soon.

colebaileygit:
> Hey @colebaileygit, could you please resolve all the conversations that have been addressed? It would be great to merge this soon.

Sorry for disappearing, had a European style vacation and getting back into things now. Will take some time to revisit the PR this week

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/file_watcher.py
colebaileygit: Sounds sensible. Would this be a simple refactor/enhancment we can add directly to this PR? It might be easier for me to review and test a first reference implementation if you can help provide that. I'm definitely not an expert in python thread handling

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/file_watcher.py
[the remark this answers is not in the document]
colebaileygit: I will mark this as closed for this PR based on the comments. @Kavindu-Dodan could you move the discussion somewhere else if you want to continue it outside the scope of this?

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/file_watcher.py
colebaileygit: Basically this:
> By setting them as daemon threads, you can let them run and forget about them, and when your program quits, any daemon threads are killed automatically.

https://stackoverflow.com/questions/190010/daemon-threads-explanation

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/config.py
```diff
@@ -42,6 +44,14 @@ def __init__(  # noqa: PLR0913
             env_or_default("FLAGD_TLS", False, cast=str_to_bool) if tls is None else tls
         )
         self.timeout = 5 if timeout is None else timeout
+        self.retry_backoff_seconds: float = (
+            float(env_or_default("FLAGD_RETRY_BACKOFF_SECONDS", 2.0))
```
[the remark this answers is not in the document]
colebaileygit: Yes, but then it gives me this:

Screenshot 2024-06-20 at 21 09 12

After fighting with mypy a bit I realised that pulling the cast outside of the function is the cleanest way to have clean typing. I would instead suggest refactoring to remove this helper function everywhere and just use the python libraries directly. This abstraction doesn't add too much if you ask me:

```
self.retry_backoff_seconds: float = (
    env_or_default("FLAGD_RETRY_BACKOFF_SECONDS", 2.0, cast=float)
    if retry_backoff_seconds is None
    else retry_backoff_seconds
)
```
vs
```
self.retry_backoff_seconds: float = (
    float(os.environ.get("FLAGD_RETRY_BACKOFF_SECONDS", 2.0))
    if retry_backoff_seconds is None
    else retry_backoff_seconds
)
```

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/flags.py
colebaileygit: Toyed with this a bit, I can pass function refs instead of object refs around to decouple things better. But if we want to keep this logic in the provider itself we cannot depend on AbstractProvider and need to depend on FlagdProvider since it now has extra functions beyond the abstract protocol.

I'll commit an attempt but I think this is up to taste, let me know if you have a specific approach to suggest

colebaileygit: Check out this commit: https://github.com/open-feature/python-sdk-contrib/pull/84/commits/4b34fadf94b0796567d74d7b4bf00ca37d9be2ef

It's a bit of a compromise. Doesn't add extra listener interfaces (which I think harm the readability of the code) or any circular dependencies, but it does achieve the goal of removing the heavy dependency on `AbstractProvider` everywhere.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/proto/flagd/evaluation/v1/evaluation_pb2_grpc.py
```diff
@@ -2,7 +2,7 @@
 """Client and server classes corresponding to protobuf-defined services."""
 import grpc
 
-from flagd.evaluation.v1 import evaluation_pb2 as flagd_dot_evaluation_dot_v1_dot_evaluation__pb2
+from . import evaluation_pb2 as flagd_dot_evaluation_dot_v1_dot_evaluation__pb2
```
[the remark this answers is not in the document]
colebaileygit: Nope, I actually didn't notice that script. When I try to run it I get this file not found error, does it need to be fixed and maybe documented in the contributing guide?
> Failure: could not read file schemas/protobuf/buf.gen.python.yaml: open schemas/protobuf/buf.gen.python.yaml: no such file or directory

beeme1mr:
> Sorry for disappearing, had a European style vacation and getting back into things now. Will take some time to revisit the PR this week

No worries. I hope you had a relaxing vacation. A European-style vacation sure sounds nice😄 At any rate, can you resolve all the addressed threads and provide a short summary of what's remaining? It looks like it's close, and I would love to help get it across the finish line.

colebaileygit:
> At any rate, can you resolve all the addressed threads and provide a short summary of what's remaining? It looks like it's close, and I would love to help get it across the finish line.

@beeme1mr I've closed some threads where I provided a best effort fix, feel free to reopen @federicobond if you think they are important!

Main points still open:
- building out a better threading solution: https://github.com/open-feature/python-sdk-contrib/pull/84#discussion_r1590530505
- adding some tests: https://github.com/open-feature/python-sdk-contrib/pull/84#discussion_r1590520726, https://github.com/open-feature/python-sdk-contrib/pull/84#discussion_r1590519851

Otherwise I'd say remaining topics can be follow-up tasks that don't need to block this PR:
- fixing grpc code generation
- fixing env_or_default cast usage
- fixing mypy and absolute imports
- fixing files embedded in published package

PR #88, opened 2024-05-07
feat: implement basic functionality of OFREP provider

federicobond:


providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
```diff
@@ -1,49 +1,180 @@
...
+                json=_build_request_data(evaluation_context),
+                timeout=self.timeout,
+            )
+            response.raise_for_status()
+
+        except requests.RequestException as e:
+            self._handle_error(e)
+
+        try:
+            data = response.json()
```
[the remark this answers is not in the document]
federicobond: Correct, I am not validating the response shape at the moment.

mypy.ini
```diff
@@ -15,3 +15,6 @@ follow_imports = silent
 
 [mypy-grpc]
 ignore_missing_imports = True
+
+[mypy-requests.*]
+ignore_missing_imports = True
```
beeme1mr, line 20: Just for my own understanding, why is this needed?

providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
```diff
@@ -1,49 +1,180 @@
...
+        *,
+        headers: Optional[Dict[str, str]] = None,
+        timeout: float = 5.0,
+    ):
+        self.base_url = base_url
+        self.headers = headers
+        self.timeout = timeout
+        self.retry_after: Optional[datetime] = None
+        self.session = requests.Session()
+        self.session.headers["User-Agent"] = "OpenFeature/1.0.0"
```
beeme1mr, line 40: Where does this version come from? I wonder if it should match the library. If so, we could use Release Please to manage it for us.

beeme1mr (approved):

providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
Kavindu-Dodan: You can give the option to accept a custom header provider callback, so we don't have to hardcode these values. Also, this will solve any authentication header requirements.

See Go implementation here - https://github.com/open-feature/go-sdk-contrib/blob/main/providers/ofrep/provider.go#L71-L74

providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
```diff
@@ -1,49 +1,180 @@
...
-    ) -> FlagResolutionDetails[Union[dict, list]]: ...
+    ) -> FlagResolutionDetails[Union[dict, list]]:
+        return self._resolve(flag_key, default_value, evaluation_context)
+
+    def _resolve(
+        self,
+        flag_key: str,
+        default_value: Union[bool, str, int, float, dict, list],
+        evaluation_context: Optional[EvaluationContext] = None,
+    ) -> FlagResolutionDetails[Any]:
```
Kavindu-Dodan, line 95: When will we perform the type of the returned flag ? In both JS & Go we perform a check [1] [2]

[1] - https://github.com/open-feature/js-sdk-contrib/blob/main/libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts#L284-L286
[2] - https://github.com/open-feature/go-sdk-contrib/blob/main/providers/ofrep/internal/evaluate/flags.go#L38

Kavindu-Dodan:
Good start 🙌  

My main concern is [this](https://github.com/open-feature/python-sdk-contrib/pull/88#discussion_r1597048587)

providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
federicobond: We have a type guard [in the client](https://github.com/open-feature/python-sdk/blob/bd0bc1e2b7a28b1f1dcd50b63b61214131968925/openfeature/client.py#L440). Should we typecheck in the provider too?

providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
federicobond: Thanks! Can you expand on why you take a callback instead of a static value? Not sure I fully understand that use case.

mypy.ini
gruebel: fixed it 🙂

providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
```diff
@@ -1,49 +1,180 @@
...
+            raise ParseError(str(e)) from e
+
+        return FlagResolutionDetails(
+            value=data["value"],
+            reason=Reason[data["reason"]],
+            variant=data["variant"],
+            flag_metadata=data["metadata"],
+        )
+
+    def _handle_error(self, exception: requests.RequestException) -> None:
```
gruebel, line 127:
```suggestion
    def _handle_error(self, exception: requests.RequestException) -> NoReturn:
```
this fits better the purpose of the method, you need to import it from `typing`

providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
```diff
@@ -1,49 +1,180 @@
...
+        if evaluation_context.targeting_key:
+            data["context"]["targetingKey"] = evaluation_context.targeting_key
+        data["context"].update(evaluation_context.attributes)
+    return data
+
+
+def _parse_retry_after(retry_after: Optional[str]) -> Optional[datetime]:
+    if retry_after is None:
+        return None
+    if re.match(r"^\s*[0-9]+\s*$", retry_after):
```
gruebel, line 177:
```suggestion
    if re.match(r"^\s*[1-9][0-9]+\s*$", retry_after):
```
how about making sure, it can't be prefixed with a `0` or is `0` in general ok as value?

gruebel (commented):
nice job technically, just added some small suggestions

providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
Kavindu-Dodan: @federicobond  If we accept static value, we must deal with expired tokens. But by outsourcing it to a callback, we can expect the callback to handle refresh steps

providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
Kavindu-Dodan: Good point. Most provider implementations I have seen perform this type validation at the provider side. IMO this is better so the provider can track its own errors, telemetry and could provide better error messages.

Kavindu-Dodan: Also, this is the client spec PR for ofrep where we define a type validation from the provider implementation [1]

[1] - https://github.com/open-feature/protocol/pull/14/files#diff-e32277be0442fee2ef2a5394baa0d98437fce4a32c3b1a6bd070747d4294f86cR51

providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
federicobond: It's semantically wrong from the provider side, but it's pretty harmless as far as I can see. Should we have a special treatment for zero?

providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
federicobond: Understood, thank you!

lukas-reining:
Hey @open-feature/sdk-python-maintainers, what is missing from your side, especially @federicobond to get this released?
It would be great for us to have OFREP in Python, so if I can help to get it through, I would be happy to help.

beeme1mr:
@lukas-reining, I think it's worth getting this out there. I'll merge when the CI passes. Thanks for kicking this off, @federicobond, and sorry for the delay.

PR #91, opened 2024-06-19
feat: Change fractional custom op from percentage-based to relative weighting.

aepfli:
relates: https://github.com/open-feature/flagd/issues/1282

### DIFF
--- a/providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/custom_ops.py
+++ b/providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/custom_ops.py
@@ -31,23 +31,35 @@ def fractional(data: dict, *args: JsonLogicArg) -> typing.Optional[str]:
         logger.error("No hashKey value resolved")
         return None
 
-    hash_ratio = abs(mmh3.hash(bucket_by)) / (2**31 - 1)
-    bucket = int(hash_ratio * 100)
+    hash_ratio = abs(mmh3.hash(bucket_by)) / (2 ** 31 - 1)
+    bucket = hash_ratio * 100
 
+    totalWeight = 0
     for arg in args:
         if (
             not isinstance(arg, (tuple, list))
-            or len(arg) != 2
-            or not isinstance(arg[0], str)
-            or not isinstance(arg[1], int)
+            or len(arg) == 0
         ):
             logger.error("Fractional variant weights must be (str, int) tuple")
             return None
+
+        if (not isinstance(arg[0], str)):
+            logger.error("Fractional variant's first element isn't string")
+            return None
+        if (len(arg) >= 2):
+            if (not isinstance(arg[1], int)):
+                logger.error("Fractional variant's second element isn't int")
+                return None
+        else:
+            arg.append(1)
+
+        totalWeight += arg[1]
+
     variant_weights: typing.Tuple[typing.Tuple[str, int]] = args  # type: ignore[assignment]
 
     range_end = 0
     for variant, weight in variant_weights:
-        range_end += weight
+        range_end += weight * 100 / totalWeight
         if bucket < range_end:
             return variant
 
@@ -69,7 +81,7 @@ def ends_with(data: dict, *args: JsonLogicArg) -> typing.Optional[bool]:
 
 
 def string_comp(
-    comparator: typing.Callable[[str, str], bool], data: dict, *args: JsonLogicArg
+        comparator: typing.Callable[[str, str], bool], data: dict, *args: JsonLogicArg
 ) -> typing.Optional[bool]:
     if not args:
         logger.error("No arguments provided to string_comp operator.")


providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/custom_ops.py
```diff
@@ -69,7 +81,7 @@ def f(s1: str, s2: str) -> bool:
 
 
 def string_comp(
-    comparator: typing.Callable[[str, str], bool], data: dict, *args: JsonLogicArg
+        comparator: typing.Callable[[str, str], bool], data: dict, *args: JsonLogicArg
```
matthewelwell, line 84: The formatting should be consistent with the rest of the codebase which uses `ruff`. Check out `pre-commit` and the configuration [here](https://github.com/open-feature/python-sdk-contrib/blob/main/.pre-commit-config.yaml). 

You should be able to run `pre-commit install` in the root directory and it will prevent you from committing mal-formatted code. 

```suggestion
    comparator: typing.Callable[[str, str], bool], data: dict, *args: JsonLogicArg
```

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/custom_ops.py
```diff
@@ -31,23 +31,35 @@ def fractional(data: dict, *args: JsonLogicArg) -> typing.Optional[str]:
         logger.error("No hashKey value resolved")
         return None
 
-    hash_ratio = abs(mmh3.hash(bucket_by)) / (2**31 - 1)
-    bucket = int(hash_ratio * 100)
+    hash_ratio = abs(mmh3.hash(bucket_by)) / (2 ** 31 - 1)
+    bucket = hash_ratio * 100
 
+    totalWeight = 0
```
matthewelwell, line 37: Python variables should always be snake_case not camelCase. 

```suggestion
    total_weight = 0
```

matthewelwell (changes requested):
I've added a few styling related comments, but this PR will also need additional test coverage. 

I also find the logic quite difficult to follow. Can we try to avoid multiple nested if statements? And can we perhaps make the error logs more descriptive about why they are error cases?

aepfli:
> I've added a few styling related comments, but this PR will also need additional test coverage.
> 
> I also find the logic quite difficult to follow. Can we try to avoid multiple nested if statements? And can we perhaps make the error logs more descriptive about why they are error cases?

I am working on adding further tests to the test harness. But i can also try to add here more tests. python, as you might have guessed, is not my strong suit. but i will try to adapt

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/custom_ops.py
aepfli: thank you, i was struggeling with getting the linting to run - should be better now

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/custom_ops.py
aepfli: my inner java dev resurfaced here ;)

aepfli:
the python sdk only logs and error and continues its targeting evaluation - compared to the other sdks, eg [java](https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Fractional.java#L104-L123) or [js](https://github.com/open-feature/js-sdk-contrib/blob/main/libs/shared/flagd-core/src/lib/targeting/fractional.ts#L74-L98) - is this a unintended difference between those, should we also unify this, and instead of logging an error raise an exception?

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/custom_ops.py
```diff
@@ -10,6 +10,12 @@
 logger = logging.getLogger("openfeature.contrib")
 
 
+class Fraction:
+    weight: int = 1
+    variant: str
+    pass
```
Kavindu-Dodan, line 16: nit - is `pass` needed here ?

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/custom_ops.py
```diff
@@ -32,28 +38,50 @@ def fractional(data: dict, *args: JsonLogicArg) -> typing.Optional[str]:
...
         if bucket < range_end:
-            return variant
+            return fraction.variant
 
     return None
 
 
+def __parse_fraction(arg: JsonLogicArg) -> typing.Optional[Fraction]:
+    if not isinstance(arg, (tuple, list)) or len(arg) == 0:
+        logger.error("Fractional variant weights must be (str, int) tuple")
```
Kavindu-Dodan, line 62: nit - since we support `{[string], [string]}`, could improve the error message here

logger.error("Fractional variant weights must be (str, int) tuple or a [string] list")

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/custom_ops.py
```diff
@@ -32,28 +38,50 @@ def fractional(data: dict, *args: JsonLogicArg) -> typing.Optional[str]:
...
-            return None
-    variant_weights: typing.Tuple[typing.Tuple[str, int]] = args  # type: ignore[assignment]
-
-    range_end = 0
-    for variant, weight in variant_weights:
-        range_end += weight
+        fraction = __parse_fraction(arg)
+        if fraction:
+            fractions.append(fraction)
+            total_weight += fraction.weight
```
Kavindu-Dodan, line 49: In both go & java, an error in parsing makes the full evaluation invalid. To be consistent, I think we should handle the parsing errors and propagate them.

Kavindu-Dodan:
> the python sdk only logs and error and continues its targeting evaluation - compared to the other sdks, eg [java](https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Fractional.java#L104-L123) or [js](https://github.com/open-feature/js-sdk-contrib/blob/main/libs/shared/flagd-core/src/lib/targeting/fractional.ts#L74-L98) - is this a unintended difference between those, should we also unify this, and instead of logging an error raise an exception?

+1 for this proposal. I had the same doubt when reviewing the PR. To be consistent, yes we should raise sceptions

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/custom_ops.py
```diff
@@ -32,28 +38,50 @@ def fractional(data: dict, *args: JsonLogicArg) -> typing.Optional[str]:
...
+        range_end += fraction.weight * 100 / total_weight
         if bucket < range_end:
-            return variant
+            return fraction.variant
 
     return None
 
 
+def __parse_fraction(arg: JsonLogicArg) -> typing.Optional[Fraction]:
+    if not isinstance(arg, (tuple, list)) or len(arg) == 0:
```
gruebel, line 61:
```suggestion
    if not isinstance(arg, (tuple, list)) or not arg:
```
a truthy/falsey test is faster, than checking the length, if you only want to know it is empty or not.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/custom_ops.py
gruebel, line 16:
```suggestion
@dataclass
class Fraction:
    variant: str
    weight: int = 1
```
please use a `dataclass` for such a case and fields with a default value go to the bottom

gruebel (commented):
nice work, added just some minor comments

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/custom_ops.py
aepfli: i will extract this into an own bug, as this is definitely worth an own pr and an own merge commit

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/custom_ops.py
aepfli: i dont thing so, i will adapt the code also based on @gruebel 's suggestion using dataclass

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/custom_ops.py
aepfli: extracted into #92

Kavindu-Dodan: Thank you @aepfli

Kavindu-Dodan (approved):

gruebel (approved):
thanks, looks good 🍻

gruebel:
@beeme1mr do you have any idea about the failing uploads to codecov? I thought we fixed that issue or is it something new?

beeme1mr:
> @beeme1mr do you have any idea about the failing uploads to codecov? I thought we fixed that issue or is it something new?

I'm not sure what's wrong. The logs indicate that the codecov secret isn't set but the configuration looks correct and it was working last week.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/custom_ops.py
```diff
@@ -32,28 +39,51 @@ def fractional(data: dict, *args: JsonLogicArg) -> typing.Optional[str]:
...
+    for fraction in fractions:
+        range_end += fraction.weight * 100 / total_weight
         if bucket < range_end:
-            return variant
+            return fraction.variant
 
     return None
 
 
+def __parse_fraction(arg: JsonLogicArg) -> typing.Optional[Fraction]:
```
matthewelwell, line 61:
```suggestion
def _parse_fraction(arg: JsonLogicArg) -> typing.Optional[Fraction]:
```

Convention in python is to use a single leading underscore for private methods (as I think is the aim here), double leading underscores are used for name mangling and (afaik) are only relevant for class methods.

matthewelwell (changes requested):

aepfli:
Any help for codecov is highly appreciated - it seems that this is the only reason the builds are failing

beeme1mr:
@aepfli could you please rebase? I think I've resolved the CodeCov issue with [this PR](https://github.com/open-feature/python-sdk-contrib/pull/93).

aepfli:
done i am not sure if anything else is missing here ;) but please let me know

PR #95, opened 2024-09-09
chore: add Python 3.13 support

gruebel:


gruebel:
ok, this needs to wait for gRPC to support 3.13 https://github.com/grpc/grpc/issues/34922

beeme1mr:
@gruebel, it looks like gRPC now supports Python 3.13!

gruebel:
Nice, giving it a try.

federicobond (approved):

federicobond:
Will this work without `allow-prereleases: true` now?

gruebel:
Yeah, it should work without now.

beeme1mr (approved):

.github/workflows/build.yml
```diff
@@ -37,6 +37,7 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           cache: "pip"
+          allow-prereleases: true
```
beeme1mr, line 40:
```suggestion
```

Let's see if we can remove this.

PR #101, opened 2024-11-09
test: Add Testcontainers and Gherkin execution for our test-harness

aepfli:
This pull request adds a gherkin runner with test containers

This approach:
- uses Testcontainers to spin up the docker images
- runs the test suites dynamically based on the spec and test-harness
- 
Issues detected during this implementation
- #102 - hence the steps within the `test_rpc.py` are overwritten
- No Events implemented - hence the steps within the `test_rpc.py` are overwritten


aepfli:
None of the evaluations are working because I get a GRPC.unavailable error when I try to resolve them. This might be related to the open issues. @toddbaert @beeme1mr fyi

beeme1mr:
@aepfli what issue are you referring to?

aepfli:
> @aepfli what issue are you referring to?

nvm I found the issue, the docker container took longer than expected to be available - now there are only 4 test cases missing, and we run the basic test suite as we do for js for the rpc provider - reconnect is something which  i will do at a later stage

aepfli:
> Could you please add a section in the contributing doc that explains how to run the e2e tests?

So far, the command has not changed for execution. It runs with `hatch run test`, like before  - if we want to split this, I would extract this into something separate. My Python tooling knowledge is not in a state, where I can easily do this

providers/openfeature-provider-flagd/pyproject.toml
```diff
@@ -37,6 +37,9 @@ dependencies = [
   "coverage[toml]>=6.5",
   "pytest",
   "pytest-bdd",
+  "testcontainers",
+  "asserts",
+  "grpcio-health-checking==1.60.0",
```
aepfli, line 42: needed to pin this to some version, as there was an ` Detected incompatible Protobuf Gencode/Runtime versions when loading grpc_health/v1/health.proto: gencode 5.27.2 runtime 5.27.1. Runtime version cannot be older than the linked gencode version.` error

aepfli:
can somebody shortly verify, or help me with the object evaluation part  here: https://github.com/open-feature/python-sdk-contrib/blob/feca070cd03429a4d55a51b0ac525c30e34fdcde/providers/openfeature-provider-flagd/tests/e2eGherkin/steps.py#L352-L392 - i am not sure i am doing this the right way here

providers/openfeature-provider-flagd/tests/test_errors.py
```diff
@@ -27,10 +29,11 @@ def create_client(provider: FlagdProvider):
     ],
 )
 def test_file_load_errors(file_name: str):
+    path = os.path.abspath(os.path.join(os.path.dirname(__file__), "./flags/"))
```
aepfli, line 32: For me it was easier to just load one provider as a project, rather than the whole sdk, but the tests where broken, due to paths issues, with this, both versions of running the tests work

aepfli:
Ready to be reviewed @toddbaert @beeme1mr @open-feature/sdk-python-approvers @open-feature/sdk-python-maintainers

providers/openfeature-provider-flagd/tests/e2e/flagd_container.py
```diff
@@ -0,0 +1,57 @@
...
+        # First we wait for Flagd to say it's listening
+        wait_for_logs(
+            self,
+            "listening",
+            5,
+        )
+
+        time.sleep(1)
+        # # Second we use the GRPC health check endpoint
+        # with grpc.insecure_channel(host + ":" + port) as channel:
```
beeme1mr: Can this be removed?

providers/openfeature-provider-flagd/tests/e2e/test_in-process-file.py
```diff
@@ -0,0 +1,81 @@
...
+)
+def provider_changed_was_executed():
+    assert_true(True)
+    # TODO: DELETE AFTER IMPLEMENTATION OF EVENTS FOR RPC
+
+
+@then(parsers.cfparse('the event details must indicate "{flag_name}" was altered'))
+def flag_was_changed():
+    assert_true(True)
+    # TODO: DELETE AFTER IMPLEMENTATION OF EVENTS FOR RPC
```
beeme1mr, line 74: Is it possible to skip the test instead?

providers/openfeature-provider-flagd/tests/e2e/flagd_container.py
aepfli: I reenabled the health check already - I do think it is a great way to check if the container is ready. I just need to figure out, how to do this with the sync testbed. But that is anyways a follow up

providers/openfeature-provider-flagd/tests/e2e/test_in-process-file.py
aepfli: There is no way of skipping scenarios if you load all of them at once - I think it is better to auto-load all of them rather than defining each test manually. When adding them manually, we add the possible human error of not adding a test, which might give us the false confidence that everything is tested based on gherkin. now I am seeing at least which features are missing somehow (still not ideal, I agree here).

providers/openfeature-provider-flagd/tests/e2e/flagd_container.py
```diff
@@ -0,0 +1,59 @@
...
+
+import grpc
+from grpc_health.v1 import health_pb2, health_pb2_grpc
+from testcontainers.core.container import DockerContainer
+from testcontainers.core.waiting_utils import wait_container_is_ready, wait_for_logs
+
+HEALTH_CHECK = 8014
+
+
+class FlagDContainer(DockerContainer):
```
toddbaert, line 11:
```suggestion
class FlagdContainer(DockerContainer):
```

providers/openfeature-provider-flagd/tests/e2e/flagd_container.py
```diff
@@ -0,0 +1,59 @@
...
+        self,
+        image: str = "ghcr.io/open-feature/flagd-testbed:v0.5.13",
+        port: int = 8013,
+        **kwargs,
+    ) -> None:
+        super().__init__(image, **kwargs)
+        self.port = port
+        self.with_exposed_ports(self.port, HEALTH_CHECK)
+
+    def start(self) -> "FlagDContainer":
```
toddbaert, line 22:
```suggestion
    def start(self) -> "FlagdContainer":
```

toddbaert:
> > `flagD` -> `flagd` but otherwise approved, and very grateful! 🙏
> 
> renamed - but we always pronounce it "flag D" else it looks more like "flagged" - just my few cents

https://flagd.dev/reference/naming/

PR #109, opened 2024-11-22
build(flagd): auto generate proto files from schema

aepfli:
autogenerating protobuf files is crucial - hence that, i applied a hatch plugin to build them automatically

The downside is, that this plugin is fairly new https://github.com/nanoporetech/hatch-protobuf - but from my check it currently looks safe to be used.

What happened in this PR:
- moved all submodules to an own directory (important for having nice paths with the protobuf generation)
- changed the test targets to proper hatch configuration, instead of running `hatch run test` it is now `hatch test` or `hatch test -c` if you want coverage
- to work with the repo we currently need to run `hatch build` first - i am investigating how to best solve this see https://github.com/pypa/hatch/discussions/1824


gruebel (commented):
nice work 🍻 

I'm actually not so worried about the plugin being new or will become unmaintained over time. I checked the code and worst case we can just copy it over or maintain our own fork of it.

Did you test the wheel creation and made sure the generated files are included?

aepfli:
> Did you test the wheel creation and made sure the generated files are included?

As I am lacking a little bit of insight/terminology. I write what I did and what I saw afterward ;)

I ran `hatch build`, which created the `dist` folder. Within the `dist` folder, I have a `.tar.gz` file, which contains the schemas in the `src` (and sadly also the test-harness folder). So, I would say this is packaged with the release. for the test-harness exclusion, I'll create my own issue/pullrequests.

aepfli:
This is still not ideal, but the proto files are now within a directory called flags within the src folder. As `flagd` is an openFeature-specific name, we should be save for the beginning. 

There is a way, to bypass this, but i think this would be an follow up (writing a script moving the files into the correct path layout, before generating the files) - also running `hatch build` before hand is not ideal, so maybe we still want to checkin the generated files

.github/workflows/build.yml
```diff
@@ -38,16 +38,15 @@ jobs:
...
       - name: Install hatch
         run: pip install hatch
 
+      - name: Building first to generate files
+        run: hatch build
+        working-directory: ${{ matrix.package }}
+
       - name: Test with pytest
-        run: hatch run cov
+        run: hatch test -c
```
toddbaert, line 49: Do we still get coverage with this?

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
```diff
@@ -13,10 +13,13 @@
     TypeMismatchError,
 )
 from openfeature.flag_evaluation import FlagResolutionDetails
+from openfeature.schemas.protobuf.flagd.evaluation.v1 import (  # type:ignore[import-not-found]
```
toddbaert, line 16:
```suggestion
from openfeature.schemas.protobuf.flagd.evaluation.v1 import (
```

Removing this gives me no problems locally. I can run `hatch build` and `hatch test` without issue.

.github/workflows/build.yml
aepfli: yes

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
aepfli: because you have run `hatch build` if somebody edits documentation, or only commits something, ruff will complain, that those are not found

aepfli: try `hatch clean` modify a file and try to commit it, normally typechecking should kick in, and prevent you from commiting

gruebel:
ok, I updated my previous PR #115 with @aepfli new changes and now it works without needing the `type:ignore` comments. I created a dedicated `hatch env` for `mypy` (which I will also do for the other 2 packages) with its own config and dependencies.

Either you can pull my changes into your branch or we merge my


fyi, nice workaround with moving the schemas to an `openfeature` subfolder!

Issue #55, opened 2024-02-23
[flagd-RPC] implement caching


toddbaert:
Implement flagd RPC caching as per [spec](https://flagd.dev/reference/specifications/rpc-providers/#caching). At a high level, "STATIC" flag values can be cached in an LRU cache

- [ ] identify adequate LRU cache lib or implement one
- [ ] use cache as per [spec](https://flagd.dev/reference/specifications/rpc-providers/#caching)
- [ ] tests

PR #110, opened 2024-11-23
feat(flagd-rpc): add caching 

aepfli:
Adding caching to the RPC resolver

closes: #55


providers/openfeature-provider-flagd/tests/e2e/config.feature
```diff
@@ -0,0 +1,87 @@
+Feature: Configuration Test
```
aepfli, line 1: I know this is a copy of the other test, but generally, I do think it might be worth adding this test to the harness to ensure all providers have the same config and support the same values. (even if they might not need this config currently)

aepfli: Maybe also adapting it, to be more feature specific

providers/openfeature-provider-flagd/tests/e2e/test_config.py
```diff
@@ -0,0 +1,90 @@
...
+        'the option "{option}" of type "{type_info}" should have the value "{value}"',
+    )
+)
+def check_option_value(option, value, type_info, config):
+    value = type_cast[type_info](value)
+    value = value if value != "null" else None
+    assert_equal(config.__getattribute__(camel_to_snake(option)), value)
+
+
+if sys.version_info >= (3, 9):
```
aepfli, line 89: this does not work on python 3.8

providers/openfeature-provider-flagd/tests/e2e/config.feature
aepfli: https://github.com/open-feature/flagd-testbed/pull/156

providers/openfeature-provider-flagd/tests/e2e/test_config.py
beeme1mr: Why isn't Python 3.8 supported? What will happen if someone uses that version?

aepfli: It is just for this test - somehow the markers don't work with python 3.8, hence i can not load this scenario, as it is not filtering properly

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
```diff
@@ -57,24 +58,40 @@ def __init__(
...
             f"{config.host}:{config.port}",
-            options=(("grpc.keepalive_time_ms", config.keep_alive_time),),
+            options=(("grpc.keepalive_time_ms", config.keep_alive),),
         )
         stub = evaluation_pb2_grpc.ServiceStub(channel)
         return stub, channel
 
     def initialize(self, evaluation_context: EvaluationContext) -> None:
         self.connect()
+        self.retry_backoff_seconds = 0.1
```
warber, line 81: Pls check if this is needed

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
```diff
@@ -57,24 +58,40 @@ def __init__(
...
+        self.connected = False
```
warber, line 82: this should be moved before we try to connect (line 80)

aepfli: thank you, this was a merge error, it is resolved now

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
```diff
@@ -249,12 +266,17 @@ def _resolve(  # noqa: PLR0915
...
         # Got a valid flag and valid type. Return it.
-        return FlagResolutionDetails(
+        result = FlagResolutionDetails(
             value=value,
             reason=response.reason,
             variant=response.variant,
         )
 
+        if response.reason == Reason.STATIC and self._cache is not None:
+            self._cache.insert(flag_key, result)
```
toddbaert, line 276: We need to only use the cache if the stream connection is active; and flush it when the stream disconnects.

toddbaert: See: https://flagd.dev/reference/specifications/providers/#cache-invalidation

aepfli: done, each time we generate a new stub and channel i will clear the cache

providers/openfeature-provider-flagd/pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
   "panzi-json-logic>=1.0.1",
   "semver>=3,<4",
   "pyyaml>=6.0.1",
+  "cachebox"
```
beeme1mr, line 27: Are there any concerns about this dependency? It seems to be Rust-based. Will that be an issue for consumers?

beeme1mr (approved):
Nice job! I'm just wondering about the caching dependency but it doesn't need to block this PR.

providers/openfeature-provider-flagd/pyproject.toml
aepfli: background from myside, why i went with cachebox:

Issue #113, opened 2024-11-24
coverage calculation seems to be off for pull requests


aepfli:
Somehow the coverage calculation for pull requests is off, see https://github.com/open-feature/python-sdk-contrib/pull/112 - I did not change the code, but I improved the coverage by some percent. 

We should investigate to have a better overview of what is changing between pull requests and how the coverage improved.

aepfli:
seems like there is no update on main coverage since #38 - this is the last info on codecov for the main branch, so I assume it might have happened afterwards

PR #114, opened 2024-11-24
build: readd build for the main branch(#113)

aepfli:


toddbaert (approved):

gruebel (approved):
Let's give it a try. If some problems pop, we can try to fix them or revert the change again.

PR #115, opened 2024-11-24
build(flagd): auto generate proto files from schema

gruebel:
## This PR

- Superseeds #109 @aepfli sadly I couldn't extend your originial PR and there a bit too many changes needed to make work


providers/openfeature-provider-flagd/pyproject.toml
```diff
@@ -57,6 +57,11 @@ cov = [
   "cov-report",
 ]
 
+[tool.hatch.build.hooks.protobuf]
+dependencies = ["hatch-protobuf"]
+proto_paths = ["schemas"]
+output_path = "src/openfeature/contrib/provider/flagd"
```
aepfli, line 63: I put the schema directly into the root of src, because of problems with package imports of the generated files - I see you're affected by the same problem. I did not find a proper solution fastly, and thought this is a good compromise

.pre-commit-config.yaml
```diff
@@ -29,4 +29,4 @@ repos:
           - mmh3
           - semver
           - panzi-json-logic
-        exclude: proto|tests
+        exclude: providers/openfeature-provider-flagd|tests
```
aepfli, line 32: I'm sorry, but I'm not really familiar with the tooling, so I will most likely ask some questions out of curiosity. Why do we want to exclude the provider from mypy? especially as we add it to the type checking in the build.yml

ruff.toml
```diff
@@ -3,7 +3,7 @@ exclude = [
     ".venv",
     "__pycache__",
     "venv",
-    "providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/proto/*"
+    "providers/openfeature-provider-flagd/src/schemas/**"
```
aepfli, line 6: i think this is now not in sync with our specified path, and also needs an update

mypy.ini
```diff
@@ -1,5 +1,5 @@
 [mypy]
-files = hooks,providers
+files = hooks,providers/openfeature-provider-ofrep
```
aepfli, line 2: [curiosity] why do we add here the ofrep provider, and not the flagd provider, should we maybe add both?

aepfli (commented):
I left some comments about some tradeoffs I made - I hope this helps; my reviews are still not the most valuable, especially with the build toolchain.  - and I am fine if this supersedes my pr. I just want to get this sdk forward and a little bit more in sync with the java version feature wise

.pre-commit-config.yaml
gruebel: because I'm moving away from a central `mypy` approach to a per package. the other ones will follow after this PR 🙂

mypy.ini
gruebel: same reason as before, moving away from a central approach

gruebel:
> I left some comments about some tradeoffs I made - I hope this helps; my reviews are still not the most valuable, especially with the build toolchain. - and I am fine if this supersedes my pr. I just want to get this sdk forward and a little bit more in sync with the java version feature wise

after doing this change I now remember why we have the shell script, I sadly have to revert this part. The big problem is we need the generated protobuf files inside the provider namespace, otherwise thet dangle somewhere outside and can conflict with any package just named `schemas`. the script did one import thing changing the import statement in the protobuf files to point to the different path.

.pre-commit-config.yaml
aepfli: Understood, but I would suggest splitting this into a separate and dedicated PR - Depending on the discussions, it can easily be that one or the other topic in this pr might lead to some controversy, and hence, one or the other improvement will be blocked by something unrelated. Furthermore adds this complexity to a topic, and for others it might be hard to understand what is now related to actually schema migration topic, and what are build improvements.

aepfli:
https://github.com/open-feature/python-sdk-contrib/pull/109/commits/107b8452ad7da13a7ece0085f652bd157651fea0

with this commit, i changed the generator to mypy-protobuf - which allowed my to specify a proper batch without issues, and custom manipulation of the protobuf files

aepfli (approved):

toddbaert (approved):

toddbaert:
Thanks!

PR #116, opened 2024-11-27
build: setup mypy for each package independently

gruebel:
## This PR

- also added OFREP provider to build CI workflow


beeme1mr (approved):
Nice, thanks!

PR #117, opened 2024-12-06
feat(flagd-rpc): adding grace attempts

aepfli:
adding grace attempts

Note:
The configuration changes are not breaking because we're passing them from the provider to the configuration. Hence that we're save


providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/config.py
```diff
@@ -19,10 +20,13 @@ class CacheType(Enum):
 DEFAULT_HOST = "localhost"
 DEFAULT_KEEP_ALIVE = 0
 DEFAULT_OFFLINE_SOURCE_PATH: typing.Optional[str] = None
+DEFAULT_OFFLINE_POLL_MS = 5000
```
aepfli, line 23: i added this to be compliant with the config gherkin test - usage of this will be fixed as soon as we touch the in-process provider

aepfli:
@toddbaert @open-feature/sdk-python-approvers @open-feature/sdk-python-maintainers @beeme1mr @warber @chrfwow a review would be highly appreciated

aepfli:
> Please mark this PR as breaking and include a list of the breaking changes in the PR description. I'm fine making the change to bring the Python implementation into spec compliance.

What is breaking?

beeme1mr:
Never mind, nothing public was changed.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
```diff
@@ -146,15 +148,37 @@ def listen(self) -> None:
...
+            retry_delay = min(1.1 * retry_delay, self.retry_backoff_max_seconds)
+        return retry_delay
+
+    def handle_error(self, retry_counter: int, retry_delay: float) -> None:
+        if retry_counter == self.retry_grace_attempts:
+            if self.cache:
+                self.cache.clear()
             self.emit_provider_error(
                 ProviderEventDetails(
                     message=f"gRPC sync disconnected, reconnecting in {retry_delay}s",
```
chrfwow, line 172: Don't we increase `retry_delay` in `handle_retry()`, so we wouldn't print the correct value here?

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
```diff
@@ -146,15 +148,37 @@ def listen(self) -> None:
                 )
 
             self.connected = False
+            self.handle_error(retry_counter, retry_delay)
+
+            retry_delay = self.handle_retry(retry_counter, retry_delay)
+
+            retry_counter = retry_counter + 1
```
chrfwow, line 155: Just for my understanding. The `retry_counter `seems to be incremented (and the error and retry handling exectued) in every loop iteration, even if the connection succeeds, or am I missing something?

aepfli: You are right. `handle_error` is a bad name for what I am actually doing. Two things are hidden behind it. 

1. if we hit our grace_attempts with the `error_counter` we emit an error event
2. if our immediate reconnect did not work, we're emitting a stale event on the first time.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
aepfli: We increment the `retry_delay` after the sleep, so in the next iteration, it will be increased. hence that the order of execution ensures the right values. But for clarity, I'll reorder the methods.

chrfwow (approved):

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
```diff
@@ -146,15 +148,37 @@ def listen(self) -> None:
                 )
 
             self.connected = False
+            self.on_connection_error(retry_counter, retry_delay)
+
+            retry_delay = self.handle_retry(retry_counter, retry_delay)
```
gruebel, line 153: just an idea. why not change `retry_counter` and `retry_delay` to private instance attributes? then you don't have to pass them in and return and can instead update them directly.

gruebel (commented):
nice work 🍻

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
aepfli: Clarity is an integral part of software development. If we put this on the class level, I am hiding important information, as the manipulation and influence of those properties are not fully transparent. Those are connection-specific properties handled and modified only during this method and submethods that are part of this logic. Hiding them on the instance or not showing their implications might be a nice shortcut for a smaller method footprint. Still, the clarity and connection of these variables to this method would be lost. Furthermore, they might also be used or seen as properties of the `resolve` logic. Hence, we should keep them close to their use case.

gruebel: if you want to go this route, then you should extract `self.retry_backoff_max_seconds` from `handle_retry()` and make it a class method. they are tightly coupled and should be treated in the same way. `handle_retry()` is just an extracted logic snippet without being reused anywhere. furthermore the incrementation of the `retry_counter` is outside of `handle_retry()` function, which contradicts your statement, because it is not clear why this is done separately and not directly inside the method.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
```diff
@@ -37,30 +37,32 @@
...
         self.emit_provider_configuration_changed = emit_provider_configuration_changed
         self.cache: typing.Optional[BaseCacheImpl] = (
             LRUCache(maxsize=self.config.max_cache_size)
-            if self.config.cache_type == CacheType.LRU
+            if self.config.cache == CacheType.LRU
             else None
         )
         self.stub, self.channel = self._create_stub()
         self.retry_backoff_seconds = config.retry_backoff_ms * 0.001
+        self.retry_backoff_max_seconds = config.retry_backoff_ms * 0.001
```
gruebel, line 62: why are there now 2 attributes with the same value? did you mean to use `config.retry_backoff_max_ms` instead?

aepfli: Nice catch thank you

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
aepfli: First off, I really appreciate the time and effort you put into reviewing pull requests—it’s clear you care about the quality of the work and the success of our project. That said, I wanted to share some thoughts on how we might make our collaboration on reviews a bit smoother and more effective.

Sometimes, the way your comments are phrased can come across as more directive or critical than collaborative. I know that’s likely not your intention, but it can make it a bit challenging to engage in a back-and-forth discussion. For example, it sometimes feels like suggestions are presented without much context, making it hard to fully understand your perspective or have a productive debate. Additionally, when there are differing opinions, it can feel like there’s more emphasis on following your suggestion than on finding a shared solution together.

I think we’re working on something really cool here, and it would be awesome if we could approach pull request reviews as more of a collaborative effort. To help with that, I’d recommend taking a look at [this blog post](https://mtlynch.io/human-code-reviews-1/), which has some great tips for making review comments more constructive and easier to follow.

I truly value your feedback, and I hope this helps make the review process feel more like a partnership. Let me know what you think, and thanks again for all the work you put into this!

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
aepfli: fixed config property usage in https://github.com/open-feature/python-sdk-contrib/pull/117/commits/38cf9b99cf04e1b9194f1738f7272fa1710db9b1

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
toddbaert: TBH the scope of this variables is not something I find that important, so neither of these refactors are a blocker for me. I can see both sides of this discussion.

As far as I can see, nothing in this PR violates the flagd spec.

toddbaert (approved):
Looks consistent with the spec to me. I don't have any concerns.

PR #118, opened 2024-12-11
feat: attempts with connection improvements

aepfli:
Based on https://github.com/open-feature/flagd/issues/1472 we're improving our reconnection logic.

We're using GRPC, and GRPC has already offers a reconnection with a backoff. 
Within this PR:
- we're moving to the reconnect mechanism of GRPC
- we're utilizing `Wait-for-ready` to wait for our stream to have a connection again
- we're migrating to a timer-based error approach (as we can't distinguish the number of connection attempts anymore)
- Adding a shutdown for the loop in the filewatcher


aepfli:
blocked by https://github.com/open-feature/flagd/pull/1478 and https://github.com/open-feature/flagd-testbed/pull/172

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
```diff
@@ -105,81 +112,87 @@ def connect(self) -> None:
                 "Blocking init finished before data synced. Consider increasing startup deadline to avoid inconsistent evaluations."
             )
 
+    def monitor(self) -> None:
+        def state_change_callback(new_state: ChannelConnectivity) -> None:
```
gruebel, line 116: can you move the inner function to the class level? I always prefer not have inner functions, if possible. too many times I had to extract them later on, because of strange side effects.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
```diff
@@ -105,81 +112,87 @@ def connect(self) -> None:
...
+                        name="FlagdGrpcServiceWorkerThread",
+                    )
+                    self.thread.start()
+
+                if self.timer and self.timer.is_alive():
+                    logger.debug("gRPC error timer expired")
+                    self.timer.cancel()
+
+            elif new_state == ChannelConnectivity.TRANSIENT_FAILURE:
+                # this is the failed reonnect attempt so we are going into stale
```
gruebel, line 132:
```suggestion
                # this is the failed reconnect attempt so we are going into stale
```

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/config.py
```diff
@@ -26,7 +26,7 @@ class CacheType(Enum):
 DEFAULT_RESOLVER_TYPE = ResolverType.RPC
 DEFAULT_RETRY_BACKOFF = 1000
 DEFAULT_RETRY_BACKOFF_MAX = 120000
-DEFAULT_RETRY_GRACE_ATTEMPTS = 5
+DEFAULT_RETRY_GRACE_PERIOD = 5
```
gruebel, line 29: what kind of unit is it? can you add a short comment?

gruebel (commented):
@aepfli let me know, when the blocker is resolved and it is good to merge

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
aepfli: thank you, extracted into a "private" method

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/config.py
aepfli: Thank you added time unit to default, and to env var - maybe we need to rethink the spec.

@toddbaert wdyt should we add a unit to the spec var name?

gruebel (approved):

gruebel:
@aepfli just to make sure, is it ready to merge now? 😅

aepfli:
> @aepfli just to make sure, is it ready to merge now? 😅 

Yes I think it is ready. The tests are also more stable ;)

PR #120, opened 2024-12-17
feat(flagd): use test-harness version number for integration tests

aepfli:


providers/openfeature-provider-flagd/tests/e2e/flagd_container.py
```diff
@@ -11,11 +12,15 @@
...
-        image: str = "ghcr.io/open-feature/flagd-testbed:v0.5.15",
+        image: str = "ghcr.io/open-feature/flagd-testbed",
         port: int = 8013,
         **kwargs,
     ) -> None:
-        super().__init__(image, **kwargs)
+        my_path = os.path.abspath(os.path.dirname(__file__))
+        path = os.path.join(my_path, "../../openfeature/test-harness/version.txt")
+        with open(path) as file:
+            data = file.read().rstrip()
```
gruebel, line 22: I recommend to use `Path` from `pathlib` here, it makes file handling much easier. There are couple of different ways to get to the path you want, so feel free to check it out
ex.
```suggestion
        path = Path(__file__).parents[2] / "openfeature/test-harness/version.txt"
        data = path.read_tex().rstrip()
```
you can also keep `../../` but then I recommend to invoke `.resolve()` once.

gruebel (commented):
are the failing tests related to the upgrade to version `0.5.18`?

providers/openfeature-provider-flagd/tests/e2e/flagd_container.py
aepfli: uh nice, that is a way cleaner api - thank you

aepfli:
> are the failing tests related to the upgrade to version `0.5.18`?

It seems like there is something flaky here. I'll check, but I am already working on a different approach for tests with an unstable connection. We don't need an image that is regularly disconnecting, as we're using test containers; we can just kill the container and be faster with our tests. Additionally this should grant us more stability in the tests, as we're in control

gruebel (approved):

PR #142, opened 2025-01-03
fix: Allow configuring ofrep provider requests to api at base …

### LINKED ISSUE #141: Using `urljoin` wth an absolute path in the OFREPProvider prevents connection to self-hosted APIs with OFREP API not at base / path
**Description:** Operations teams often host feature flagging services themselves rather than using a SaaS solution. In this instance they put their api under a path other than the base `/` path. The **_expected behaviour_** when using the OFREP Provider would be that I pass my base_url (i.e. `https://mydomain.com/tools/featureflags`) and all subsequent requests would be made using the defined base_url.

The _**actual behaviour**_ is that the `urljoin("https://mydomain.com/tools/featureflags", f"/ofrep/v1/evaluate/flags/{flag_key}")` strips the path from the base_url and produces a request to `https://mydomain.com/ofrep/v1/evaluate/flags/{flag_key}`. This is because the second param in urljoin is being passed as an absolute path that starts with "/"

**Proposed Solution**: Convert the second arg of urljoin to a relative path and add logic to ensure that the first args ends with /

```python
            evaluation_base_url = self.base_url if self.base_url.endswith("/") else f"{self.base_url}/"
            response = self.session.post(
                urljoin(evaluation_base_url, f"ofrep/v1/evaluate/flags/{flag_key}"),
                json=_build_request_data(evaluation_context),
                timeout=self.timeout,
                headers=self.headers_factory() if self.headers_factory else None,
            )
```

atmask:
## This PR

This PR updates the use of `urljoin` in the OFREP Provider so that api's at a base path other than "/" are supported.

### Related Issues

Fixes #141

### DIFF
--- a/providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
+++ b/providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
@@ -107,6 +107,12 @@ class OFREPProvider(AbstractProvider):
             FlagType.OBJECT, flag_key, default_value, evaluation_context
         )
 
+    def _get_ofrep_api_url(self, api_version: str = "v1") -> str:
+        ofrep_base_url = (
+            self.base_url if self.base_url.endswith("/") else f"{self.base_url}/"
+        )
+        return urljoin(ofrep_base_url, f"ofrep/{api_version}/")
+
     def _resolve(
         self,
         flag_type: FlagType,
@@ -124,7 +130,7 @@ class OFREPProvider(AbstractProvider):
 
         try:
             response = self.session.post(
-                urljoin(self.base_url, f"/ofrep/v1/evaluate/flags/{flag_key}"),
+                urljoin(self._get_ofrep_api_url(), f"evaluate/flags/{flag_key}"),
                 json=_build_request_data(evaluation_context),
                 timeout=self.timeout,
                 headers=self.headers_factory() if self.headers_factory else None,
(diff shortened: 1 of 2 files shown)


providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
```diff
@@ -124,7 +130,7 @@ def _resolve(
 
         try:
             response = self.session.post(
-                urljoin(self.base_url, f"/ofrep/v1/evaluate/flags/{flag_key}"),
+                urljoin(self._get_ofrep_api_url(), f"evaluate/flags/{flag_key}"),
```
aepfli, line 133: Based on this stackoverflow: https://stackoverflow.com/a/1793282/3708208, urljoin interprets a string starting with a slash as a path on the server root. Maybe we don't need the method at all, but we need to change the second parameter of the urljoin.

```suggestion
                urljoin(self.base_url, f"ofrep/v1/evaluate/flags/{flag_key}"),
```

aepfli (approved):
Thank you nice catch, and this looks really good to me - i am just wondering, if my suggestion, might also do the trick, with lesser code change :)

providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
atmask: I had thought about this as well. The only issue is that, if a user provides a base_url which does not end in a trailing slash then the last part of their provided base_url will get dropped by `urljoin()`. For example, `urljoin("https://localhost:8080/tools/featureFlags", "ofrep/v1/keys")` becomes `'https://localhost:8080/tools/ofrep/v1/keys'`. 

We can add in some logic to ensure that the base_url ends with a slash and have it all in `_resolve`. I just broke it out into a separate function so that it was easier to create the unit tests for validating that the ofrep provider's url parsing behaviour remains consistent and doesn't get broken over time

aepfli: Ah understood!

beeme1mr (approved):

PR #143, opened 2025-01-03
feat: Add Flipt provider

### LINKED ISSUE #89: Add support for Flipt provider
i've been reading about Openfeature and noticed that some of the languages already have a Flipt provider but the Python SDK is missing one.

Example of the Golang one: https://github.com/open-feature/go-sdk-contrib/tree/main/providers/flipt

atmask:
## Overview

This PR implements the Flipt provider by wrapping around the OFREP Provider and leveraging Flipt's OFREP api

## Discussion Points:
- Should we support passing the namespace in the constructor this way or just leave it up to users to do it themselves via the generic headers_factory (like they would need to for authentication)

### Related Issues
Fixes #89 

### How to test

Unit Tests: From the base of the package run `hatch test`

Local Usage: Go to the base of the new packager (i.e. `provider/openfeature-provider-flipt`), run `pip install -e .`, and then open a python repl with `python3`:
```python
from openfeature import api
from openfeature.contrib.provider.flipt import FliptProvider
api.set_provider(FliptProvider(base_url="<flipt instance>", namespace="<your namespace>"))
client = api.get_client()
client.get_boolean_value("<your feature flag key>", True)
```

### DIFF
--- a/providers/openfeature-provider-flipt/src/openfeature/contrib/provider/flipt/__init__.py
+++ b/providers/openfeature-provider-flipt/src/openfeature/contrib/provider/flipt/__init__.py
@@ -1,0 +1,57 @@
+from typing import Callable, Dict, Optional
+
+from openfeature.contrib.provider.ofrep import OFREPProvider
+from openfeature.provider import Metadata
+
+# Only export the FliptProvider class symbol form the package
+__all__ = ["FliptProvider"]
+
+
+class FliptProvider(OFREPProvider):
+    """A provider for Flipt feature flags service that extends the OFREP Provider for interacting with Flipt's OFREP API"""
+
+    def __init__(
+        self,
+        base_url: str,
+        namespace: str,
+        *,
+        headers_factory: Optional[Callable[[], Dict[str, str]]] = None,
+        timeout: float = 5.0,
+    ):
+        """Override the OFREPProvider constructor to add a namespace parameter"""
+
+        # Build a headers factory function that includes the Flipt namespace header
+        headers_factory = self._resolve_header_factory(namespace, headers_factory)
+        super().__init__(base_url, headers_factory=headers_factory, timeout=timeout)
+
+    def _resolve_header_factory(
+        self, namespace: str, headers_factory: Optional[Callable[[], Dict[str, str]]]
+    ) -> Callable[[], Dict[str, str]]:
+        """
+        Resolves and returns a headers factory callable that includes the "X-Flipt-Namespace" header.
+
+        If a headers factory is provided, it will be called and its headers will be merged with the
+        "X-Flipt-Namespace" header. If no headers factory is provided, a new factory will be created
+        that only includes the "X-Flipt-Namespace" header.
+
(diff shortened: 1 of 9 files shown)


markphelps (approved):
lgtm! just curious if this gives end users much over the generic Python OFREP impl? Is the end goal of OFREP to make it easier to implement provider specific sdks or to be 'the' SDK for all providers that support it?

cc @beeme1mr

providers/openfeature-provider-flipt/src/openfeature/contrib/provider/flipt/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+            Callable[[], Dict[str, str]]: A callable that returns a dictionary of headers including
+            the "X-Flipt-Namespace" header.
+        """
+        if headers_factory is None:
+            headers = {"X-Flipt-Namespace": namespace}
+        else:
+            headers = {
+                **headers_factory(),
+                "X-Flipt-Namespace": namespace,
+            }
```
atmask, line 52: Something we might want to consider is that this will result in the output of the user's header's factory being called once and the output being merged into the dict here. That dict will always be returned without the user's header's factory being called again. This could be fine it you are okay with it but it is a change from the current implementation where the user's headers_factory is called on every feature flag evaluation in the OFREP provider.

I think I could turn this into a closure around the user's headers_factory callable if that is something we might be concerned about.

beeme1mr:
I'm fine with the OFREP provider being extended to make it work more seamlessly. It's really up to you and Mark. I believe Flipt also has an SDK that supports in-process evaluation. That would likely be a more performant option.

atmask:
@beeme1mr  kk, sounds good. I'll leave that ask up to @markphelps to give direction on. I had thought of wrapping around the existing Python SDK for Flipt but OFREP seemed quicker and easier option to get support out (I'd like to be able to consume this api). I can re-work it but it'll just take some more time to re-write the implementation

markphelps:
@atmask lgtm! thank you!! we do have an inprocess evaluator like @beeme1mr mentioned, although I think it would be classified as a client provider in the OF world. the only thing to note is that Flipt added OFREP support in Flipt v1.47.0, so these endpoints wont exist in any versions of Flipt before this

beeme1mr (approved):
Could you please registry this provider in the release please manifest and config? Once that's in place, we should be good to release. Thanks!

- https://github.com/open-feature/python-sdk-contrib/blob/main/.release-please-manifest.json
- https://github.com/open-feature/python-sdk-contrib/blob/main/release-please-config.json

atmask:
@beeme1mr Yes! Just pushed the config updates

atmask:
@beeme1mr What blockers remain for getting this merged and releasing? Is release to pypi automated atm?

beeme1mr:
> What blockers remain for getting this merged and releasing? Is release to pypi automated atm?

Sorry, the blocker is me. I'll take care of this ASAP.

PR #159, opened 2025-01-21
ci: update release process

beeme1mr:
## This PR

- removes the Python container so the pypi release process works 🤞

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,8 +45,6 @@ jobs:
     permissions:
       # IMPORTANT: this permission is mandatory for trusted publishing to pypi
       id-token: write
-    # container:
-    #   image: "python:3.13"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -45,14 +45,16 @@ jobs:
...
       id-token: write
-    container:
-      image: "python:3.13"
 
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
         with:
           submodules: recursive
 
+      - uses: actions/setup-python@v5
```
gruebel, line 54:
```suggestion
      - uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5
```

gruebel (approved):

Issue #163, opened 2025-01-22
[Feat] Add support for Flipt Client-Side evaluation model


atmask:
**Description:** Flipt cloud only supports client-side evaluation models. Self-hosted flipt instances support both. To provider a wider coverage of open-feature for end-users there should be support for both evaluation models.

Currently the flipt provider wraps the ofrep provider which is fine for evaluation against a self-hosted instance, however, this implementation will need to wrap the official client side sdk: https://github.com/flipt-io/flipt-client-sdks/tree/main/flipt-client-python

@markphelps tagging you for visibility

atmask:
I'll start working on adding the support for this. Should be done in not too long. I just started tonight. It will new python provider: `openfeature-provider-flipt-client`

PR #179, opened 2025-02-05
feat: [WIP] Add the open feature flipt client. Tests still required

atmask:
## This PR

**Background:**
This PR adds a new provider to support flipt-client evaluation. This provider more flexibility to end users looking to consume from the Flipt and it allows OpenFeature to be used directly alongside Flipt Cloud which only supports the Client side sdk atm (self-hosted supports both client and server side sdks).

**Implementation**
- Flipt Client Provider is an adapter around the Official Flipt Client SDK
- The Evalation API is implemented with strict typing meaning
  - default_value provided has a type assertion done for the expected evaluation type (str, int, float, dict, list)
  - the type of the  returned response from the official sdk is checked. If TYPE_MISMATCH occurs then the default is returned and appropriate error code and reason are in the evaluation details
- I have tried my best to map FliptEvaluationReasons that I see in (https://github.com/flipt-io/flipt-client-sdks/blob/7f6be0aaee75938e5b235344451d898d583c0fe9/flipt-evaluation/src/models/flipt.rs#L144) and the js providers mapping of flipt Reasons (https://github.com/open-feature/js-sdk-contrib/blob/main/libs/providers/flipt-web/src/lib/models.ts#L17) over to the generic OpenFeature reasons where possible but the Official Flipt client SDK does not define an Enum of reasons and the OpenFeature Reson's field support a Reson enum or Str _so where mapping cannot be done cleanly the reason string returned from the flipt client sdk is passed through transparently_

### Related Issues
#163

### DIFF
--- a/providers/openfeature-provider-flipt-client/README.md
+++ b/providers/openfeature-provider-flipt-client/README.md
@@ -1,0 +1,37 @@
+# OpenFeature Flipt Client-Side Provider
+
+This provider is designed to support client side evaluation of feature flags for Flipt feature flag provider.
+
+-----
+
+## Table of Contents
+
+- [Installation](#installation)
+- [License](#license)
+
+## Installation
+
+```console
+pip install openfeature-provider-flipt-client
+```
+
+## Configuration
+
+```python
+from openfeature.contrib.provider.flipt_client import (
+    FliptClientProvider,
+)
+
+FliptClientProvider(
+    base_url="",
+    Client
+)
+
+
+```
+
+
+
+## License
+
+Apache 2.0 - See [LICENSE](./LICENSE) for more information.
--- a/providers/openfeature-provider-flipt-client/src/openfeature/contrib/provider/flipt-client/__init__.py
+++ b/providers/openfeature-provider-flipt-client/src/openfeature/contrib/provider/flipt-client/__init__.py
@@ -1,0 +1,235 @@
+from typing import Optional, Union, Any
+
+from enum import Enum
+from openfeature.provider import AbstractProvider, Metadata
+from flipt_client.models import (
+    FetchMode, 
+    AuthenticationStrategy, 
+    ClientTokenAuthentication,
+    JWTAuthentication,
+    ClientOptions
+)
+from flipt_client import FliptEvaluationClient
+
+from openfeature.evaluation_context import EvaluationContext
+from openfeature.exception import ErrorCode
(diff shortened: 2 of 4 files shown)


providers/openfeature-provider-flipt-client/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+FliptClientProvider(
+    base_url="",
+    Client
+)
+
+
+```
+
+
```
markphelps, line 33: we may want to put a note in the README that this provider will only work on platforms that use Glibc as the C library. Also it will only work for the OS'es mentioned [here](https://github.com/flipt-io/flipt-client-sdks?tab=readme-ov-file#ffi) because we are using FFI under the hood for the evaluation engine clientside: https://github.com/flipt-io/flipt-client-sdks?tab=readme-ov-file#ffi

markphelps (commented):
lgtm so far! one comment about adding a potential compatability note in the readme

thank you @atmask !!

providers/openfeature-provider-flipt-client/src/openfeature/contrib/provider/flipt-client/__init__.py
```diff
@@ -0,0 +1,235 @@
...
+                update_interval=update_interval,
+                authentication=authentication,
+                reference=reference,
+                fetch_mode=fetch_mode
+            )
+        )
+    
+    def get_metadata(self) -> Metadata:
+        return Metadata(
+            name='OpenFeature Flipt Client-Side Provider',
```
beeme1mr, line 48:
```suggestion
            name='Flipt',
```

I'd recommend keeping this as short as possible since it may be used in telemetry and logs.

providers/openfeature-provider-flipt-client/src/openfeature/contrib/provider/flipt-client/__init__.py
```diff
@@ -0,0 +1,235 @@
...
+                flag_key=flag_key,
+                entity_id=self._get_entity_id(evaluation_context),
+                context=evaluation_context.attributes
+            )
+        except Exception as e:
+            return FlagResolutionDetails(
+                value=default_value,
+                reason=Reason.ERROR,
+                error_message=str(e),
+                error_code=None ## Explicitly set to None. No clear mapping atm from Flipt error to OpenFeature error codes
```
beeme1mr, line 78: Please use the `General` error code if there isn't a more appropriate one.

PR #211, opened 2025-03-12
chore(flagd): Add sync metadata disabled

cupofcat:
## This PR
- Adds option which disables GetMetadata call

### Related Issues
- https://github.com/open-feature/flagd/issues/1584

### Notes
`GetMetadata` call was recently added in https://github.com/open-feature/python-sdk-contrib/pull/195. This is potentially breaking for server implementations that don't support metadata. As such, this PR is a temporary stop-gap for such implementations, and it will be removed when metadata is deprecated with the completion of the above issue.

There's no need to implement this in all languages, we'll do it at maintainer discretion according to user need.

This is similar to recently added option in the Java SDK: https://github.com/open-feature/java-sdk-contrib/pull/1267

CC: @toddbaert @aepfli @gruebel

### DIFF
--- a/providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/config.py
+++ b/providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/config.py
@@ -253,7 +253,7 @@ class Config:
         # TODO: remove the metadata call entirely after https://github.com/open-feature/flagd/issues/1584
         # This is a temporary stop-gap solutions to support servers that don't implement sync.GetMetadata
         # (see: https://buf.build/open-feature/flagd/docs/main:flagd.sync.v1#flagd.sync.v1.FlagSyncService.GetMetadata).
-        # Using this option sisables call to sync.GetMetadata
+        # Using this option disables call to sync.GetMetadata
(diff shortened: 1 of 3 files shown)


beeme1mr (approved):

toddbaert (approved):

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/config.py
```diff
@@ -248,3 +249,11 @@ def __init__(  # noqa: PLR0913
         )
 
         self.channel_credentials = channel_credentials
+
+        # TODO: remove the metadata call entirely after https://github.com/open-feature/flagd/issues/1584
+        # This is a temporary stop-gap solutions to support servers that don't implement sync.GetMetadata
+        # (see: https://buf.build/open-feature/flagd/docs/main:flagd.sync.v1#flagd.sync.v1.FlagSyncService.GetMetadata).
+        # Using this option sisables call to sync.GetMetadata
```
gruebel, line 256:
```suggestion
        # Using this option disables call to sync.GetMetadata
```

gruebel (approved):
Thanks, just one small thing

PR #214, opened 2025-03-16
fix(flagd): handle falsy target values correctly

### LINKED ISSUE #213: [Potential Bug] FILE resolver type falls back to default value
## Description

When using the flagd provider with the FILE resolver type in Python, the feature flag evaluation sometimes returns the `default_value` instead of the expected variant from our flags.json. For example, if you call .get_boolean_value() multiple times on the same flag name but with different default values, the client falls back to the default instead of returning the configured variant.

## Sample code
Sample `flags.json`:

```
{
    "$schema": "https://flagd.dev/schema/v0/flags.json",
    "flags": {
        "enable-debug-logs": {
            "defaultVariant": "custom_key",
            "state": "ENABLED",
            "targeting": {
                "if": [
                    {
                        "in": [
                            {
                                "var": "instance"
                            },
                            [
                                "placeholder",
                                "custom_key"
                            ]
                        ]
                    },
                    {
                        "var": "instance"
                    },
                    null
                ]
            },
            "variants": {
                "placeholder": true,
                "custom_key": false
            }
        }
    }
}
```

Sample code:

```
from openfeature import api
from openfeature.contrib.provider.flagd import FlagdProvider
from openfeature.contrib.provider.flagd.config import ResolverType
from openfeature.evaluation_context import EvaluationContext

evaluation_ctx = EvaluationContext(
    targeting_key="instance",
    attributes={
        "instance": "custom_key",
    },
)

api.set_evaluation_context(evaluation_ctx)
api.set_provider(
    FlagdProvider(
        resolver_type=ResolverType.FILE,
        offline_flag_source_path="flags.json",
    )
)

client = api.get_client()

print(client.get_boolean_value("enable-debug-logs", default_value=False))
print(client.get_boolean_value("enable-debug-logs", default_value=True))
```

You should see the output:
```
False
True
```
which means the `default_value` was returned. 

## Environment details
```
Python 3.11.11
openfeature-provider-flagd==0.2.0
openfeature_sdk==0.8.0
```

AdityaVallabh:
## This PR

Prevents erroneous exception when dealing with falsy values.

### Related Issues

Fixes #213 

### Notes

N/A

### Follow-up Tasks

N/A

### How to test

Added a bunch of UTs.


providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/in_process.py
```diff
@@ -121,8 +121,6 @@ def _resolve(
             )
 
         variant, value = flag.get_variant(variant)
-        if not value:
```
aepfli, line 124: i like your solution. It is clean, and offers what we want, but - and this is something i am not sure about with python - we might want to move away from exceptions and use flow control aka returning rather than throwing. 

We did this already in java and in javascript - because exception do have an overhead versus returning here a `FlagResolutionDetails`-Object with an errorstatus - see [here for java](https://github.com/open-feature/java-sdk/pull/1095). With this in mind, i think it would be better to keep this check here. But as i said my python knowledge is too limited to actually state that this is also more costly to use exception versus flow. Maybe @gruebel has better insights for us.

aepfli (commented):
thank you for your great contribution. This solves your issue as described, but i am not sure, we want to continue in the future with exceptions anyways for handling our flow. Hence i wrote a little comment, and hope some of the Python-masters can help and guide us.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/in_process.py
AdityaVallabh: I understand. How does this look: https://github.com/AdityaVallabh/python-sdk-contrib/commit/88a8aa659fb186bc775b3f2e72537cea7f6726ca?

aepfli: i think changing this line to 
```
if value is None:
```

could also do the trick

AdityaVallabh: That would work for most zero-values except "None" itself. What if I actually want to return a "None" target value? It would throw an exception instead.
Specifically, this testcase would fail: https://github.com/open-feature/python-sdk-contrib/blob/cd6b7e7bc3513ff180726699882524bf5983717a/providers/openfeature-provider-flagd/tests/test_in_process.py#L147

AdityaVallabh: Which is why I simply moved the check inside and modified the condition without touching the exception logic.
We were already throwing an exception before and continue to do so after this PR too.
If we want to move away from exceptions, we could probably do that in a separate PR in various other places as well.

aepfli: `null` or `None` violate the JSON Schema, it has to be an object. I am not sure, we want to support invalid configurations. Hence this, we should opt for simple and easy to read code, rather complex handling, with additional return values.
image

AdityaVallabh: Got it, sure will move that line back here.

aepfli (approved):

beeme1mr (approved):
Thanks!

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/in_process.py
aepfli: thank you, again for helping and supporting us. this contribution and finding was really really important, and it showed us, a spot which we missed in our gherkin test suite

gruebel: typically raising an error is the way to go and for the scope of this PR good enough. But I think we should move this error and the one above it into the `FlagResolutionDetails` as mentioned.

gruebel (approved):
Thanks 🍻 if you need help with the lints let me know.

PR #280, opened 2025-07-15
fix(flagd): Add upperbound to protobuf version to reduce potential conflict with other version

aepfli:
Currently, our release artifact is built with Protobuf version 6. This can cause trouble if another version of Protobuf is used in a project. Based on https://protobuf.dev/support/cross-version-runtime-guarantee/#python, we should be able to support backward compatibility on a higher level by binding it to this range.

As I am not the best person within the Python universe, please share your insights with me if this definition is too lax and we should just use version 4 at all!

FYI: @gruebel


toddbaert (approved):
@aepfli is the title and description still accurate?

gruebel (approved):

PR #282, opened 2025-07-16
feat(flagd): use json schema validation instead of custom validation

aepfli:
changed the code to rely on json schema instead of custom validation, should be more error proof for the future

### DIFF
--- a/providers/openfeature-provider-flagd/pyproject.toml
+++ b/providers/openfeature-provider-flagd/pyproject.toml
@@ -26,7 +26,6 @@ dependencies = [
   "pyyaml>=6.0.1",
   "cachebox",
   "jsonschema",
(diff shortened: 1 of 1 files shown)


providers/openfeature-provider-flagd/pyproject.toml
```diff
@@ -24,7 +24,9 @@ dependencies = [
   "panzi-json-logic>=1.0.1",
   "semver>=3,<4",
   "pyyaml>=6.0.1",
-  "cachebox"
+  "cachebox",
+  "jsonschema",
+  "types-jsonschema"
```
gruebel, line 29: The types dependency goes under the `[tool.hatch.envs.mypy]` block

gruebel (approved):
Nice work

PR #285, opened 2025-07-21
fix(flagd): adjust flagd to updated error codes

leakonvalinka:
Fixing the inconsistencies regarding error codes for edge cases, as described in [#1679](https://github.com/open-feature/flagd/issues/1679)


providers/openfeature-provider-flagd/tests/e2e/inprocess/flags/allFlags.json
```diff
@@ -0,0 +1,787 @@
+{
```
toddbaert, line 1: @leakonvalinka @aepfli is this file supposed to be committed? This looks like the e2e flags we provide from the test-harness. I'm not sure why we'd need a copy here.

gruebel (approved):
looks good 🍻

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/in_process.py
```diff
@@ -128,30 +128,55 @@ def _resolve(
...
+    metadata: typing.Mapping[str, typing.Union[float, int, str, bool]],
+    reason: Reason,
+) -> FlagResolutionDetails:
+    variant, value = flag.default
+    if variant is None:
+        return FlagResolutionDetails(
+            value,
+            variant=variant,
+            reason=Reason.ERROR,
+            error_code=ErrorCode.FLAG_NOT_FOUND,
```
aepfli, line 175: I compared this to the other methods, and there is a special error message used ""Flag '" + key+ "' has no default variant defined, will use code default"" maybe we should adapt the tests to add the message to be in sync in all of them? wdyt? @leakonvalinka

PR #293, opened 2025-08-02
feat: migrate to uv

### LINKED ISSUE #283: Migrate to uv
Now that [uv](https://github.com/astral-sh/uv) is available and works more reliably than our current build setup, we should consider migrating to it. This would especially help new contributors, as it would reduce the project setup pains, and make it easier and faster to start working.

We already migrated the [python sdk](https://github.com/open-feature/python-sdk/issues/502), so we should be able to do it here as well.

gruebel:
## This PR

- also all projects are using `uv_build` as build backend, except `flagd`, because there is no build hook support yet

### Related Issues

Fixes #283 

### Follow-up Tasks

- ~I'll also check out, if we can leverage `uv` workspace without `release-please` support (not part of this PR)~ workspace added


aepfli:
Thank you soooooo much! I really enjoy working with uv and I thinkt his can make our live soo much easier.

Do you think it is worth to add a pyproject.toml in the root and use workspaces? I am not sure if it adds value, but basically that is how all the other contrib repos are working. (At least that what I understood ;) )

gruebel:
> Thank you soooooo much! I really enjoy working with uv and I thinkt his can make our live soo much easier.
> 
> Do you think it is worth to add a pyproject.toml in the root and use workspaces? I am not sure if it adds value, but basically that is how all the other contrib repos are working. (At least that what I understood ;) )

@aepfli ok, I think I made it work. Only challenge was `flipt` type checking setup, but that works now too 🙂

aepfli (approved):

PR #304, opened 2025-09-25
return default dict if not meta provided

sysradium:
When using with flipt I observed that since it does not always return `metadata` key in the payload we end up having this:
```python
 File "/private/tmp/.venv/lib/python3.13/site-packages/openfeature/contrib/provider/ofrep/__init__.py", line 155, in _resolve
    flag_metadata=data["metadata"],
                  ~~~~^^^^^^^^^^^^
KeyError: 'metadata'
```
I suggest just to have a fallback to a empty dict?


gruebel:
@sysradium change looks good. as far as I can tell it also reflects the specs to return an empty record. can you check the DCO part, it looks like you forgot to sign-off the last commit.

PR #311, opened 2025-10-28
Update OFREP provider to use httpx and support async

### LINKED ISSUE #310: Support async in OFREP provider
Right now the OFREP provider uses requests, which means that it can only support sync APIs.

Would there be any willingness to move to using the `httpx` library for http requests rather than `requests`? It's API compatible so should be a straightforward port, and it would then become trivial to add support for async requests.

I'm willing to open a PR for this but I'll prioritize it more highly if I get explicitly confirmation whether you are open to it.

dmontagu:
## This PR

Updates the OFREP provider to use httpx and have proper async support.

I have not yet added tests etc. to this PR, because I wanted to get feedback on it before doing more work. But I have used this version of the provider code internally and it has worked fine, so I at least wanted to share.

### Related Issues

Closes #310 if merged.

### Follow-up Tasks

Need to update pyproject.toml, add tests, etc. Probably will take a decent effort so I don't want to undertake it if there's not consensus this is a good approach.

I'd be potentially willing to port all existing usage of requests to httpx if there was interest in expanding beyond just this OFREP provider (and with that, adding async support for all providers in this package). I'm not sure what the right replacement for requests-mock is with httpx (which I can see is what is used in tests) but I'm sure something reasonable exists.

### How to test

This won't work yet unless you manually (uv) pip install httpx. I'm mostly looking to get confirmation of interest in the approach / replacing requests with httpx at this time.

### DIFF
--- a/providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
+++ b/providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
@@ -1,12 +1,13 @@
+import asyncio
 import re
 from collections.abc import Mapping, Sequence
 from datetime import datetime, timedelta, timezone
 from email.utils import parsedate_to_datetime
+from json import JSONDecodeError
 from typing import Any, Callable, NoReturn, Optional, Union
 from urllib.parse import urljoin
 
-import requests
-from requests.exceptions import JSONDecodeError
+import httpx
 
 from openfeature.evaluation_context import EvaluationContext
 from openfeature.exception import (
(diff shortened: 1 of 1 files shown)


providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
```diff
@@ -55,7 +56,23 @@ def __init__(
...
+
+        self.client = httpx.Client()
+        self.client_async = httpx.AsyncClient()
+        self._client_async_is_entered = False
+
+    def initialize(self, evaluation_context: EvaluationContext) -> None:
+        self.client.__enter__()
+
+    def shutdown(self) -> None:
+        self.client.__exit__(None, None, None)
```
gruebel: why not just call `.close()` on it

providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
```diff
@@ -171,11 +239,66 @@ def _resolve(
...
+            await self.client_async.__aenter__()
+            self._client_async_is_entered = True
+
+        now = datetime.now(timezone.utc)
+        if self.retry_after and now <= self.retry_after:
+            raise GeneralError(
+                f"OFREP evaluation paused due to TooManyRequests until {self.retry_after}"
+            )
+        elif self.retry_after:
+            self.retry_after = None
```
gruebel: duplicated code, please create a small private method for it

providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
```diff
@@ -205,6 +328,10 @@ def _handle_error(self, exception: requests.RequestException) -> NoReturn:
 
         raise OpenFeatureError(error_code, error_details) from exception
 
+    def __del__(self):
+        # Ensure clients get cleaned up
+        self.shutdown()
```
gruebel: not sure, if this is really needed, but it is better to use `weakref` for this.

gruebel (commented):
thanks for the contribution 🍻  I don't mind switching to `httpx`, but please add some test case(s) for the async variant and you also need to adjust the `pyproject.toml`

PR #346, opened 2026-02-20
fix(ofrep): handle 401, 403, 404 and 5xx errors explicitly and add tests

thomaspoignant:
## Description

**What was the problem?**
- The OFREP provider did not explicitly handle 401, 403, 404, or generic 5xx responses before attempting to parse JSON. This could cause parse errors when servers returned non-JSON error bodies.
- For 404, responses with invalid JSON would fail instead of falling back to the response text.
- We had parse errors when it was not a parsing issue.

**How is it resolved?**
- Add explicit handling for 401 (unauthorized), 403 (forbidden), 404 (flag not found with JSON/text fallback), and status codes > 400 before the generic JSON parsing path.
- These cases now raise appropriate OpenFeatureError or FlagNotFoundError without attempting JSON parse.

**How can we test?**
- Run `pytest tests/test_provider.py -v` in `providers/openfeature-provider-ofrep/`.

## Checklist

- [x] I have followed the [CONTRIBUTING](../CONTRIBUTING.md) guidelines.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.


providers/openfeature-provider-ofrep/src/openfeature/contrib/provider/ofrep/__init__.py
```diff
@@ -176,24 +178,19 @@ def _handle_error(self, exception: requests.RequestException) -> NoReturn:
...
         try:
             data = response.json()
         except JSONDecodeError:
             raise ParseError(str(exception)) from exception
 
-        error_code = ErrorCode(data["errorCode"])
+        try:
+            error_code = ErrorCode(data["errorCode"])
+        except ValueError:
+            error_code = ErrorCode.GENERAL
```
gruebel: can you add a warn log message here? you would need to create a logger, like this
```
logger = logging.getLogger("openfeature.contrib")
```

PR #377, opened 2026-04-08
feat(flagd): extract evaluator into api, core, and testkit packages

aepfli:
## Summary

Mirrors the Java SDK contrib architecture ([PR #1696](https://github.com/open-feature/java-sdk-contrib/pull/1696), [PR #1742](https://github.com/open-feature/java-sdk-contrib/pull/1742)) by extracting the flagd evaluation logic into three independent packages:

- **`openfeature-flagd-api`** (`tools/openfeature-flagd-api/`): `Evaluator` Protocol defining the contract for flag evaluation, so others can implement their own evaluator
- **`openfeature-flagd-core`** (`tools/openfeature-flagd-core/`): Reference implementation (`FlagdCore`) with targeting engine and custom operators (fractional v2, sem_ver, starts_with, ends_with)
- **`openfeature-flagd-api-testkit`** (`tools/openfeature-flagd-api-testkit/`): Compliance test suite bundling gherkin feature files from the test-harness `evaluator/` directory, with pytest-bdd step definitions — installable as a package so custom evaluator implementations can run the same compliance suite

### Provider refactoring
- `InProcessResolver` now delegates evaluation to `FlagdCore` via an adapter pattern
- Old modules (`flags.py`, `targeting.py`, `custom_ops.py`) are thin re-exports from core for backward compatibility
- Connectors (`FileWatcher`, `GrpcWatcher`) remain unchanged
- No changes to gRPC resolvers, config, or other provider functionality

### Fractional bucketing
- `flagd-core` implements the **v2 fractional algorithm** (unsigned hash, integer arithmetic with `(hash * totalWeight) >> 32`)
- Includes `MAX_WEIGHT_SUM` overflow guard, negative weight clamping, explicit bool-as-weight rejection
- v1 fractional tests are deselected since the implementation is v2

### CI & release
- Added `tools/*` packages to the build workflow matrix (lint, mypy, tests on Python 3.10–3.14)
- Added `py.typed` marker files for PEP 561 compliance
- Replaced `project.scripts` with `poethepoet` tasks to match CI conventions
- Added release-please config for all 3 new packages
- Added `tools/*` to UV workspace members

### Other changes
- Updated test-harness submodule to v3.5.0 (adds `evaluator/` directory with gherkin feature files)
- Dropped Python 3.9 support for tools packages (aligned with rest of project)
- Schemas and spec submodules kept at main (no changes)

## Test plan

- [x] `openfeature-flagd-api` unit tests: 10 passed
- [x] `openfeature-flagd-api-testkit` smoke tests: 2 passed, mypy clean
- [x] `openfeature-flagd-core` unit tests: 27 passed
- [x] `openfeature-flagd-core` e2e (testkit compliance): 85 passed, 15 deselected (fractional-v1), 0 failures
- [x] Provider unit tests: no regressions
- [x] Lint: ruff check + ruff format clean
- [x] Type checking: mypy strict clean for all 3 tools packages

### How to use the testkit (for custom evaluator implementations)

```python
# conftest.py
from openfeature.contrib.tools.flagd.testkit import load_testkit_flags
from openfeature.contrib.tools.flagd.testkit.steps import *  # noqa: F403

@pytest.fixture
def evaluator():
    core = MyCustomEvaluator()
    core.set_flags(load_testkit_flags())
    return core
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)


providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/in_process.py
```diff
@@ -1,33 +1,39 @@
...
+        emit_provider_configuration_changed: typing.Callable[
+            [ProviderEventDetails], None
+        ],
+    ):
+        self.evaluator = evaluator
+        self.emit_provider_configuration_changed = emit_provider_configuration_changed
+
+    def update(self, flags_data: dict) -> None:
+        json_str = json.dumps(flags_data)
+        changed_keys = self.evaluator.set_flags_and_get_changed_keys(json_str)
```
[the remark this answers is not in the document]
aepfli: Done — the `Evaluator` protocol and `FlagdCore` now accept `str | dict`, so `_FlagStoreAdapter` passes the dict directly. No more serialization roundtrip.

tools/openfeature-flagd-api/src/openfeature/contrib/tools/flagd/api/evaluator.py
```diff
@@ -0,0 +1,40 @@
...
+from openfeature.evaluation_context import EvaluationContext
+from openfeature.flag_evaluation import FlagResolutionDetails, FlagValueType
+
+
+class Evaluator(typing.Protocol):
+    def set_flags(self, flag_configuration_json: str) -> None: ...
+
+    def set_flags_and_get_changed_keys(
+        self, flag_configuration_json: str
+    ) -> list[str]: ...
```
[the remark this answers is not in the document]
aepfli: Done — updated the protocol to accept `str | dict[str, Any]`.

tools/openfeature-flagd-core/src/openfeature/contrib/tools/flagd/core/flagd_core.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def set_flags(self, flag_configuration_json: str) -> None:
+        with self._lock:
+            data = json.loads(flag_configuration_json)
+            self._flag_store.update(data)
+
+    def set_flags_and_get_changed_keys(self, flag_configuration_json: str) -> list[str]:
+        with self._lock:
+            data = json.loads(flag_configuration_json)
+            return self._flag_store.update(data)
```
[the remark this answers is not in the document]
aepfli: Done — `set_flags` and `set_flags_and_get_changed_keys` now accept `str | dict[str, Any]` and skip `json.loads` when a dict is passed.

tools/openfeature-flagd-core/src/openfeature/contrib/tools/flagd/core/flagd_core.py
```diff
@@ -0,0 +1,221 @@
...
+                elif not isinstance(variant, str):
+                    variant = str(variant)
+
+                if variant not in flag.variants:
+                    raise GeneralError(
+                        f"Resolved variant {variant} not in variants config."
+                    )
+
+            except ReferenceError:
+                raise ParseError(f"Invalid targeting {targeting}") from ReferenceError
```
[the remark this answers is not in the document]
aepfli: Good catch — fixed both issues: now chains from the caught exception instance `e` and logs `flag.targeting` instead of the function object.

tools/openfeature-flagd-core/src/openfeature/contrib/tools/flagd/core/model/flag_store.py
```diff
@@ -0,0 +1,53 @@
...
+        """Update flags and return list of changed flag keys."""
+        flags = flags_data.get("flags", {})
+        metadata = flags_data.get("metadata", {})
+        evaluators: dict | None = flags_data.get("$evaluators")
+        if evaluators:
+            transposed = json.dumps(flags)
+            for name, rule in evaluators.items():
+                transposed = re.sub(
+                    rf"{{\s*\"\$ref\":\s*\"{name}\"\s*}}", json.dumps(rule), transposed
+                )
```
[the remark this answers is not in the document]
aepfli: Done — added `re.escape(name)` to prevent regex injection from evaluator names with special characters.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/flags.py
```diff
@@ -1,124 +1,35 @@
...
-        self, variant_key: str | bool | None
-    ) -> tuple[str | None, typing.Any]:
-        if isinstance(variant_key, bool):
-            variant_key = str(variant_key).lower()
-
-        if not variant_key:
-            return None, None
-
-        return variant_key, self.variants.get(variant_key)
+        return changed_keys
```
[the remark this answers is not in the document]
aepfli: Done — changed the return type to `None` for backward compatibility. Also switched `flags_changed` from `list(self.flags.keys())` to `changed_keys` so the event only reports flags that actually changed, not every flag in the store.

tools/openfeature-flagd-api-testkit/src/openfeature/contrib/tools/flagd/testkit/utils.py
```diff
@@ -0,0 +1,18 @@
...
+def str2bool(v: str) -> bool:
+    return v.lower() in ("yes", "true", "t", "1")
+
+
+type_cast: dict[str, typing.Callable] = {
+    "Integer": int,
+    "Float": float,
+    "String": str,
+    "Boolean": str2bool,
+    "Object": lambda v: json.loads(v.replace('\\\\"', '"')),
```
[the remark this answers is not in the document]
aepfli: Investigated this — the `replace` is actually needed. Gherkin feature files use `\"` to escape double-quotes inside table cells. When pytest-bdd parses these, the resulting Python string contains literal backslash-quote sequences, which `json.loads` cannot parse directly. The replace call normalizes these into valid JSON before parsing. Removing it causes 2 test failures in the Object evaluation scenarios (test_resolve_basic_values and test_resolve_zero_values_statically for object flags). Added a comment to explain this.

toddbaert:
I'm reviewing this, will complete tomorrow. Thanks @aepfli

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/in_process.py
toddbaert, line 32: `set_flags_and_get_changed_keys` already accepts `dict`, so this serializes to JSON only for `FlagdCore` to immediately deserialize it. Could pass `flags_data` directly and skip the round-trip.

tools/openfeature-flagd-api/src/openfeature/contrib/tools/flagd/api/exceptions.py
```diff
@@ -0,0 +1,6 @@
+class FlagStoreError(Exception):
```
toddbaert, line 1: Is this raised anywhere? I don't see it used in `flagd-core`. If it's part of the `Evaluator` contract, it should be documented (when should implementors raise it?). If not, consider removing it.

tools/openfeature-flagd-core/src/openfeature/contrib/tools/flagd/core/flagd_core.py
```diff
@@ -0,0 +1,231 @@
...
+                json.loads(flag_configuration)
+                if isinstance(flag_configuration, str)
+                else flag_configuration
+            )
+            self._flag_store.update(data)
+
+    def set_flags_and_get_changed_keys(
+        self, flag_configuration: str | dict[str, typing.Any]
+    ) -> list[str]:
+        with self._lock:
```
toddbaert, line 82: This JSON parsing block is duplicated from `set_flags` above (lines 71-75). Consider extracting a `_parse_flag_configuration` helper.

tools/openfeature-flagd-core/src/openfeature/contrib/tools/flagd/core/flagd_core.py
```diff
@@ -0,0 +1,231 @@
...
+            metadata[key] = value
+    return metadata
+
+
+def _default_resolve(
+    flag: Flag,
+    default_value: T,
+    metadata: Mapping[str, float | int | str | bool],
+    reason: Reason,
+) -> FlagResolutionDetails:
```
toddbaert, line 47: nit: return type should be `FlagResolutionDetails[T]` (parameterized) for type safety.

toddbaert:
This mirrors the Java api/core/testkit split, which is great. One thing worth thinking through early: this creates an inter-package dependency chain (provider -> core -> api), and that has ongoing management consequences.

The provider now depends on `"openfeature-flagd-core"` with no version constraint, and core depends on `"openfeature-flagd-api"` the same way. Any time core adds or changes an API that the provider uses, you need to coordinate: bump core, release it, then bump the provider's minimum core version. Miss that, and users can end up with a core version that's missing the method the provider calls; in Python that's a silent runtime `AttributeError` (no compile-time safety net like Java).

The `tool.uv.sources` workspace resolution means dev and CI always use the local copy, so these version mismatches won't surface until users install from PyPI. You'd need a dedicated CI job installing published versions to catch this.

Some concrete suggestions:
1. Pin minimum versions now: `"openfeature-flagd-core>=0.1.0"`, `"openfeature-flagd-api>=0.1.0"`. Establish the convention that these get bumped whenever a new core/api surface is consumed.
2. Consider compatible release constraints (`~=0.1`) during 0.x to limit blast radius, since any minor bump is potentially breaking per semver.
3. In Java and [JS](https://github.com/open-feature/js-sdk-contrib/blob/main/libs/shared/flagd-core/package.json) we ended up cutting a [1.0 for core](https://github.com/open-feature/java-sdk-contrib/pull/1755) specifically so we could set a meaningful upper bound and semver contract. We could even start core at 1.0 here to avoid the 0.x ambiguity from the outset; that way the provider can declare something like `>=1.0,<2` and breaking changes are communicated clearly.

None of this is necessarily blocking, but the release coordination overhead is real (we've felt it in Java), and it's easier to set up the guardrails now than to retrofit them after a few broken releases.

toddbaert:
I pulled this locally and did a careful review. I found a few issues worth fixing I think, but fundamentally I'm close to approving. Please consider [this](https://github.com/open-feature/python-sdk-contrib/pull/377#issuecomment-4269893277) carefully... WDYT about going right to 1.0? Or going to 1.0 very quickly after merging?

aepfli:
> I pulled this locally and did a careful review. I found a few issues worth fixing I think, but fundamentally I'm close to approving. Please consider [this](https://github.com/open-feature/python-sdk-contrib/pull/377#issuecomment-4269893277) carefully... WDYT about going right to 1.0? Or going to 1.0 very quickly after merging?

I agree, I focused to much on extraction, rather than versioning, and i totally overlooked this part. Thank you for pointing out. I will follow your suggestion and improve this.

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/in_process.py
aepfli: Done — removed the `json.dumps` call and pass the dict directly. Also dropped the now-unused `import json`.

tools/openfeature-flagd-api/src/openfeature/contrib/tools/flagd/api/exceptions.py
aepfli: Good question. It's not raised by `flagd-core` today, but it's part of the public API surface (exported from `flagd-api`). Removing it would be a breaking change. Added a docstring documenting when Evaluator implementors should raise it (corrupt configuration, I/O failures during flag loading).

tools/openfeature-flagd-core/src/openfeature/contrib/tools/flagd/core/flagd_core.py
aepfli: Done — extracted a `_parse()` static method that both `set_flags` and `set_flags_and_get_changed_keys` now call.

tools/openfeature-flagd-core/src/openfeature/contrib/tools/flagd/core/flagd_core.py
aepfli: Done — changed to `FlagResolutionDetails[T]`.

toddbaert:
> > I pulled this locally and did a careful review. I found a few issues worth fixing I think, but fundamentally I'm close to approving. Please consider [this](https://github.com/open-feature/python-sdk-contrib/pull/377#issuecomment-4269893277) carefully... WDYT about going right to 1.0? Or going to 1.0 very quickly after merging?
> 
> I agree, I focused to much on extraction, rather than versioning, and i totally overlooked this part. Thank you for pointing out. I will follow your suggestion and improve this.

I suggest either making it 1.0 here, or immediately after we release and do some basic testing.

toddbaert (approved):

toddbaert:
I'll merge this today/tomorrow unless I hear objections. I think I will do the 1.0.0 of core a bit later @aepfli

PR #388, opened 2026-04-28
chore: resolve open dependabot security alerts

jonathannorris:
## Summary

- Resolved all 6 open Dependabot security alerts by bumping 4 transitive dependencies in `uv.lock`: `urllib3` (3 high-severity alerts), `mako`, `python-dotenv`, and `pygments`
- No direct dependency constraints needed updating; all vulnerable packages are transitive deps resolved purely in the lockfile

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1083,14 +1083,14 @@ wheels = [
 
 [[package]]
 name = "mako"
-version = "1.3.10"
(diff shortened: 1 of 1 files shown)


gruebel:
thanks! but I think this is already covered by #372

PR #389, opened 2026-05-08
chore: resolve open dependabot security alerts

jonathannorris:
## Summary

- Bumped `mako` to `>=1.3.12` via a `constraint-dependencies` entry in the root `pyproject.toml` to resolve CVE-2026-44307 (high severity, path traversal via backslash URI on Windows in `TemplateLookup`)
- Regenerated `uv.lock` with `mako` 1.3.11 -> 1.3.12

Fixes Dependabot alert #30.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,9 @@ dev = [
 [tool.uv]
 required-version = ">=0.10.0"
 package = false
+constraint-dependencies = [
+  "mako>=1.3.12",
+]
 
 [tool.uv.sources]
 # hooks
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -28,6 +28,9 @@ dev = [
 [tool.uv]
 required-version = ">=0.10.0"
 package = false
+constraint-dependencies = [
```
gruebel: I think this is not needed, a normal uv sync update should be enough. There is also a specialized command to only update a single package

jonathannorris: yea, good call — we shouldn't be adding `constraint-dependencies` here. will update to use `uv lock --upgrade-package mako` instead and drop that pyproject.toml change.

matthewelwell (approved):

gruebel (approved):

PR #391, opened 2026-05-10
fix(flagd): back off before stream reconnect

### LINKED ISSUE #390: fix: add explicit application-level backoff on stream reconnection
Both the RPC and in-process resolvers lack an explicit application-level delay before re-establishing the stream after an error or completion. They rely entirely on gRPC channel-level mechanisms (`wait_for_ready`, channel backoff options), which don't prevent tight loops when the stream itself errors immediately (e.g., an intervening L7 proxy returning errors on stream creation).

Per the [provider spec](https://flagd.dev/reference/specifications/providers/#stream-reconnection), providers must apply an explicit application-level delay of `retryBackoffMaxMs` before re-establishing the stream. All other provider implementations (Go, Java, JS, .NET) already do this.

**Affected resolvers:**
- `resolvers/grpc.py` (RPC)
- `process/connector/grpc_watcher.py` (in-process)

looooown2006:
## Summary

Fixes the flagd provider stream reconnect loop so both gRPC stream implementations apply an application-level delay before creating the next stream after a stream error or normal stream completion.

Changes:
- `GrpcResolver` now waits `retry_backoff_max_ms` before recreating the RPC `EventStream`.
- `GrpcWatcher` now waits `retry_backoff_max_ms` before recreating the in-process `SyncFlags` stream.
- shutdown uses an event-backed wait so shutdown can interrupt the backoff instead of blocking until the timeout expires.
- added regression coverage for RPC and in-process stream error/completion reconnect paths.

Closes #390.

## Verification

```text
uv run --frozen pytest tests/test_grpc_watcher.py tests/test_grpc_resolver.py -q
# 10 passed
```

```text
uv run --frozen pytest tests/test_config.py tests/test_errors.py tests/test_file_store.py tests/test_flagd.py tests/test_grpc_watcher.py tests/test_grpc_resolver.py tests/test_in_process.py tests/test_metadata.py tests/test_targeting.py -q
# 199 passed
```

```text
uv tool run ruff check providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/grpc_watcher.py providers/openfeature-provider-flagd/tests/test_grpc_watcher.py providers/openfeature-provider-flagd/tests/test_grpc_resolver.py
# All checks passed
```

```text
uv run --frozen poe mypy
# Success: no issues found in 26 source files
```

Note: I also tried the broader non-e2e pytest command, but the local run still initialized the Docker-backed e2e test-harness fixture and failed before running those cases because Docker Desktop's Linux engine was not running on this Windows machine.

### DIFF
--- a/providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
+++ b/providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
@@ -67,10 +67,12 @@ class GrpcResolver:
         logger.debug(self.config.fatal_status_codes)
 
         self.retry_grace_period = config.retry_grace_period
+        self.retry_backoff_max_seconds = config.retry_backoff_max_ms * 0.001
         self.streamline_deadline_seconds = config.stream_deadline_ms * 0.001
         self.deadline = config.deadline_ms * 0.001
         self.connected = False
         self._is_fatal = False
+        self._shutdown_event = threading.Event()
         self.channel = self._generate_channel(config)
         self.stub = evaluation_pb2_grpc.ServiceStub(self.channel)
         self.selector_metadata: tuple[tuple[str, str], ...] | None = (
@@ -140,6 +142,7 @@ class GrpcResolver:
 
     def shutdown(self) -> None:
         self.active = False
+        self._shutdown_event.set()
         self.channel.unsubscribe(self._state_change_callback)
         self.channel.close()
         if self.timer and self.timer.is_alive():
@@ -150,6 +153,7 @@ class GrpcResolver:
 
     def connect(self) -> None:
         self.active = True
+        self._shutdown_event.clear()
 
         # Run monitoring in a separate thread
         self.monitor_thread = threading.Thread(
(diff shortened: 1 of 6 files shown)


looooown2006:
﻿Thanks for the review. I kept this tied to `retry_backoff_max_ms` because #390 and the flagd provider spec both describe `retryBackoffMaxMs` as the explicit application-level delay before recreating the stream after an error or completion.

This is separate from gRPC's RPC-level retry behavior; the added wait only applies between one stream ending and the next stream being created. That seems to align with the issue/spec and the reconnect-loop scenario they call out.

toddbaert:
@coderabbitai /review

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/grpc.py
```diff
@@ -228,6 +232,37 @@ def _stream_call_args(self) -> GrpcMultiCallableArgs:
             call_args["metadata"] = self.selector_metadata
         return call_args
 
+    def _wait_before_reconnect(self) -> None:
+        self._shutdown_event.wait(self.retry_backoff_max_seconds)
+
+    def _handle_rpc_error(self, e: grpc.RpcError) -> bool:
```
toddbaert, line 238: Thanks for the method extractions.

toddbaert (approved):

PR #396, opened 2026-06-01
feat: make use of flagd-selector header in RPC mode

### LINKED ISSUE #319: [flagd] Document migration guidance and update samples for selector handling
### Context
Refer to open-feature/flagd#1814 and the related implementation change to selector handling across flagd services.

### Requirements
- Document migration path for users of python-sdk-contrib with respect to selector normalization, specifically calling out changes to usage of `flagd-selector` header
- Update SDK and plugin usage samples to reflect the preferred approach
- Add deprecation notice to any examples or docs that still use request body selector
- Ensure backwards compatibility documentation is included.

### Tasks
1. Update user-facing documentation and guides for selector normalization
2. Ensure SDK/plugin samples use the new preferred header
3. Add a section on maintaining backward compatibility and preparing for future major version breaking change

### Additional notes
- Work should align with documentation approaches used in core flagd
- Link to upstream issue: https://github.com/open-feature/flagd/issues/1814

toddbaert:
- The in-process resolver already sends the `flagd-selector` gRPC metadata header, but the RPC resolver did not. This PR wires it up: the RPC resolver now sends the header on every unary call and on the `EventStream` when a selector is configured, so the flagd server can serve the requested flag set.
- Brings the Python RPC resolver to parity with the Java and JS providers, which both already wire this header up (see flagd issue [#1814](https://github.com/open-feature/flagd/issues/1814)).
- Additive: behavior is unchanged when no selector is configured.

We missed the RPC mode in: https://github.com/open-feature/python-sdk-contrib/pull/320 (closed https://github.com/open-feature/python-sdk-contrib/issues/319).


suthar26 (approved):

JamieSinn (approved):
No concerns

PR #404, opened 2026-07-24
feat(flagd): support custom gRPC metadata on in-process SyncFlags

### LINKED ISSUE #403: flagd provider: allow custom gRPC metadata on the in-process SyncFlags stream
## Problem

The in-process resolver of the flagd provider opens a long-lived gRPC stream. It calls the `SyncFlags` method. The provider builds the gRPC metadata for this call. The function `_create_metadata` adds only one header. It adds the `flagd-selector` header. It adds this header only when the user sets a selector.

The provider gives the user no way to add other metadata headers. The `Config` object has no field for custom metadata. The user cannot add a header to the sync stream.

## Effect

A user cannot send an infrastructure-specific header on the sync stream. One example is a proxy timeout header. A proxy can close a stream after a default request timeout. To keep the stream open, the client must send a header such as `x-envoy-upstream-rq-timeout-ms: 0`. The current provider cannot send this header. Because of this, a proxy can close the long-lived sync stream.

## Proposed change

1. Add a `sync_metadata` option to `FlagdProvider` and to `Config`. The value is a sequence of key-value pairs (header name and header value).
2. In `_create_metadata`, add these pairs to the metadata. Keep the `flagd-selector` header. Send both the selector header and the user headers together.
3. Keep the current behavior when the user does not set `sync_metadata`. If the user sets no selector and no custom metadata, the function returns no metadata, as before.

## Compatibility

The change is backward compatible. The custom metadata is added only when the user sets `sync_metadata`. The default behavior does not change.

## Scope

The change is small. It touches `config.py`, `provider.py`, and `grpc_watcher.py`. It affects only the in-process resolver, because only that resolver uses the `SyncFlags` metadata function.

tuhinkanti:
## This PR

Closes https://github.com/open-feature/python-sdk-contrib/issues/403

Adds a `sync_metadata` option to the flagd provider so callers can send custom gRPC metadata headers on the in-process `SyncFlags` stream.

## Problem

The in-process resolver builds the gRPC metadata for the long-lived `SyncFlags` call in `_create_metadata`, which adds only the `flagd-selector` header (and only when a selector is set). There is no way for a caller to add other metadata headers, so infrastructure-specific headers cannot be sent on the sync stream.

## Change

- Add a `sync_metadata` option to `FlagdProvider` and `Config` (a sequence of `(key, value)` header pairs).
- In `_create_metadata`, append the configured pairs to the metadata, alongside the `flagd-selector` header.
- Return no metadata when neither a selector nor custom metadata is set, preserving current behavior.

## Use case

A proxy can close a stream after a default request timeout. Sending a header such as `x-envoy-upstream-rq-timeout-ms: 0` on the long-lived sync stream keeps the stream open.

## Compatibility

Backward compatible. Custom metadata is added only when `sync_metadata` is set. The default behavior does not change.

## Tests

New tests cover config passthrough and metadata construction (selector-only, custom-only, both together, and disabled).

```
$ pytest tests/test_config.py tests/test_grpc_watcher.py -k "sync_metadata or metadata" -v

tests/test_config.py::test_sync_metadata_passthrough PASSED
tests/test_grpc_watcher.py::TestGrpcWatcher::test_custom_sync_metadata_appended PASSED
tests/test_grpc_watcher.py::TestGrpcWatcher::test_custom_sync_metadata_without_selector PASSED
tests/test_grpc_watcher.py::TestGrpcWatcher::test_listen_with_sync_metadata_and_sync_context PASSED
tests/test_grpc_watcher.py::TestGrpcWatcher::test_listen_with_sync_metadata_disabled_in_config PASSED
tests/test_grpc_watcher.py::TestGrpcWatcher::test_listen_with_sync_metadata_only PASSED
tests/test_grpc_watcher.py::TestGrpcWatcher::test_selector_passed_via_both_metadata_and_body PASSED
```

Full suite for the two touched test files:

```
$ pytest tests/test_config.py tests/test_grpc_watcher.py

20 passed
```

### DIFF
--- a/providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/grpc_watcher.py
+++ b/providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/grpc_watcher.py
@@ -212,17 +212,21 @@ class GrpcWatcher(FlagStateConnector):
 
         return request_args
 
-    def _create_metadata(self) -> tuple[tuple[str, str]] | None:
+    def _create_metadata(self) -> tuple[tuple[str, str], ...] | None:
         """Create gRPC metadata headers for the request.
 
         Returns gRPC metadata as a tuples of tuples containing header key-value pairs.
         The selector is passed via the 'flagd-selector' header per flagd v0.11.0+ specification,
         while also being included in the request body for backward compatibility with older flagd versions.
+        Any user-configured ``sync_metadata`` headers are appended, allowing callers to inject
+        infrastructure-specific headers (e.g. proxy/mesh timeout overrides) on the sync stream.
         """
-        if self.selector is None:
-            return None
+        metadata: tuple[tuple[str, str], ...] = ()
+        if self.selector is not None:
+            metadata += (("flagd-selector", self.selector),)
+        metadata += self.config.sync_metadata
 
-        return (("flagd-selector", self.selector),)
+        return metadata if metadata else None
 
     def _fetch_metadata(self) -> sync_pb2.GetMetadataResponse | None:
         if self.config.sync_metadata_disabled:
--- a/providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/config.py
+++ b/providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/config.py
@@ -105,6 +105,7 @@ class Config:
         channel_credentials: grpc.ChannelCredentials | None = None,
         sync_metadata_disabled: bool | None = None,
         fatal_status_codes: list[str] | None = None,
+        sync_metadata: typing.Sequence[tuple[str, str]] | None = None,
     ):
         self.host = env_or_default(ENV_VAR_HOST, DEFAULT_HOST) if host is None else host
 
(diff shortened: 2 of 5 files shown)


toddbaert:
Thanks for this; the problem is valid, proxies killing the long-lived sync stream is a realistic issue.

My worry is that `sync_metadata` is pretty specific. Custom headers aren't really a flagd concept, and if we add this I suspect we'll keep getting "can we also set X on the channel" requests.

Since grpcio already handles this, what about exposing a gRPC **interceptor** option instead? An interceptor that adds your `x-envoy-upstream-rq-timeout-ms: 0` header covers your case and any future one, without flagd growing a new option each time. It also matches what Java and Go already do, middleware stays outside the normal flagd options.

Docs:
- https://grpc.github.io/grpc/python/grpc.html#client-side-interceptor
- https://grpc.github.io/grpc/python/grpc.html#grpc.intercept_channel

Would you be up for reworking it that way? Happy to help.

tuhinkanti:
Let me look and get back to you

providers/openfeature-provider-flagd/src/openfeature/contrib/provider/flagd/resolvers/process/connector/grpc_watcher.py
```diff
@@ -293,27 +293,37 @@ def _handle_rpc_error(self, e: grpc.RpcError) -> bool:
     def _wait_before_reconnect(self) -> None:
         self._shutdown_event.wait(self.retry_backoff_max_seconds)
 
+    def _listen_once(
```
tuhinkanti, line 296: This is to address Ruff complexity > 10

tuhinkanti:
@toddbaert Let me know what you think or if you had something else in mind

PR #416, opened 2026-09-11
feat(provider-tck): let an adopter add their own scenarios to the suite

aepfli:
Stacked on #413. Answers @toddbaert's review request on open-feature/spec#423: an adopter with provider-specific behaviour — flagd's `fractional`, a vendor rollout rule — must be able to run their own scenarios in the same backend lifecycle as the canonical suite instead of maintaining a parallel harness. Tracking: open-feature/spec#417.

## What an adopter writes

A `tck-extensions/` directory beside the test module, and step definitions in their own `conftest.py`:

```
tests/
├── conftest.py            # @then("the fractional rule splits the population")
├── test_conformance.py    # scenarios(*feature_paths())
└── tck-extensions/
    └── fractional.feature
```

No registration, no option, no new argument. pytest collects `conftest.py` itself and pytest-bdd resolves steps through the fixture system, so a step written beside the test module is already in scope for the scenarios generated into it, with the canonical step vocabulary available alongside it; `tck_state` hands it the provider the suite registered. The only thing pytest cannot find on its own is the canonical feature files, which live inside the installed distribution — `feature_paths()` returns those plus the extension directory. The call is the same with or without extensions, and `scenarios(features_path())` still works and still sees only the canonical set.

Go and JavaScript have no runtime scanning and took explicit options; Java scans the classpath for the same `tck-extensions` name. Python needs neither.

## Shadowing

A scenario's uri in the results payload is derived from where its file is: `features/` for the packaged assets and nothing else, `extensions/` for anything below a `tck-extensions` directory — the same prefix Go and JavaScript use. This is load-bearing rather than cosmetic. pytest-bdd names a feature file by its parent directory joined to its own name, so `tck-extensions/features/errors.feature` arrives as `features/errors.feature`, the uri the canonical file already occupies, and a Messages stream carries one source per uri. `tests/test_extensions.py` runs that exact layout and checks both files reach the stream under distinct uris with their own sources.

Two cases the derivation cannot rule out are refused, and no report is written for either: a file of the adopter's own that would reach the reserved `features/` prefix, and two feature files that would share one uri.

## Second commit: canonical-set guard

Separable — it guards a bypass rather than enabling anything, and dropping it leaves the extension point unaffected.

A run that executed less than the canonical set fails and writes no report, naming what is missing. A capability-gated skip counts as having run; extension scenarios do not count and cannot close a gap. `PROVIDER_TCK_PARTIAL=1` buys a green run for someone working on one scenario with `-k`, never a report — the same escape hatch Java has. It also makes scenario enumeration `trylast` so a filtered run no longer reports every deselected scenario as collected but never run.

## Verification

Local only: `build.yml` triggers on PRs targeting `main`, so a stacked PR gets no CI.

- `pytest tests`: 132 passed, 9 skipped, 2 xfailed. `ruff check`, `ruff format --check` and `mypy` clean.
- An adoption with an extension emits a report that validates against `conformance-report.schema.json` (Draft 2020-12) on the spec's `feat/provider-tck-report-schema` branch.
- Extension and canonical scenarios in one session, one provider registration, one report; an adoption with no extensions produces the same scenario identities, outcomes, sources and envelope as `scenarios(features_path())` did before.


aepfli:
The extension point moved into the base PR (#409) and the canonical-set guard into the report PR (#413), so this one has nothing left of its own.

